PR #69, opened 2017-07-14
Update XCOM tool to read current web API

markbandstra:
The XCOM web API seems to have changed sometime in the last month. This PR updates the tool so that it works again.

Two notes:
  * [`BeautifulSoup4`](https://www.crummy.com/software/BeautifulSoup/) was added as a dependency since it is needed by `pandas` to parse HTML tables. This will also ensure #63 is fixed.
  * Four webtests still fail since the Mixture query on [XCOM](https://physics.nist.gov/PhysRefData/Xcom/html/xcom1.html) seems to be broken at this time.

PR #71, opened 2017-07-27
Double the speed of NNDC tests

markbandstra:
These modifications reduce the time for running the tests from 83 seconds to 41 seconds on my machine, while maintaining the same test coverage.

I am also interested in enabling webtests for Travis CI if this PR and #69 are accepted.


jccurtis:
I observe the same 2x speed up on `faster-nndc-tests` via `pytest -m "" tests/nndc_test.py`. I also see 95% coverage before and after. Approving ...

PR #72, opened 2017-08-05
Enable webtests

markbandstra:
I would like to enable the webtests for Travis CI so that we can track problems accessing web APIs like we had with the XCOM tools. In fact, I have already enabled weekly builds for this purpose. This PR is to make sure the webtests work in Travis and to fix any outstanding webtest issues.


markbandstra:
This is probably all I want to do with this PR, but I am open to input.

markbandstra:
Could someone take a look at this? Thanks.

tests/isotope_test.py
```diff
@@ -201,7 +201,7 @@ def test_isotope_str(iso_str, sym, A, m):
 
 
 ISOTOPE_PROPERTIES = [
-    ('H-3', (3.888E8, False, None, '1/2+', 0., None, [['B-'], [100.]],)),
+    ('H-3', (3.888E8, False, None, '1/2+', 0., 14.94980957, [['B-'], [100.]],)),
```
jccurtis: Why was this `None` before? I'm unfamiliar with this test...

markbandstra: That's the mass excess, and apparently the database used to have an empty entry for it.

jccurtis:
Besides my previous comment this looks good to me.

Issue #74, opened 2017-08-24
Test error in test_isotopequantity_decays_during[Tl-208]


markbandstra:
An error has recently started cropping up in the Travis builds:

```
=================================== FAILURES ===================================
__________________ test_isotopequantity_decays_during[Tl-208] __________________
iq = <becquerel.tools.isotope_qty.IsotopeQuantity object at 0x7f2bfb390410>
    def test_isotopequantity_decays_during(iq):
        """Test IsotopeQuantity.*_during()"""
    
        if iq.half_life > 1000 * 3.156e7:
            # avoid overflow errors in test calculations
            pass
        else:
            dt_s = iq.half_life
            t0 = datetime.datetime.now()
            t1 = t0 + datetime.timedelta(seconds=dt_s)
            spec = Spectrum(np.zeros(256), start_time=t0, stop_time=t1)
    
>           assert np.isclose(iq.decays_during(spec), iq.atoms_now() / 2)
E           assert False
E            +  where False = <function isclose at 0x7f2c23534e60>(49118462.303944729, (98235824.678773701 / 2))
E            +    where <function isclose at 0x7f2c23534e60> = np.isclose
E            +    and   49118462.303944729 = <bound method IsotopeQuantity.decays_during of <becquerel.tools.isotope_qty.IsotopeQuantity object at 0x7f2bfb390410>>(<becquerel.core.spectrum.Spectrum object at 0x7f2bfa3797d0>)
E            +      where <bound method IsotopeQuantity.decays_during of <becquerel.tools.isotope_qty.IsotopeQuantity object at 0x7f2bfb390410>> = <becquerel.tools.isotope_qty.IsotopeQuantity object at 0x7f2bfb390410>.decays_during
E            +    and   98235824.678773701 = <bound method IsotopeQuantity.atoms_now of <becquerel.tools.isotope_qty.IsotopeQuantity object at 0x7f2bfb390410>>()
E            +      where <bound method IsotopeQuantity.atoms_now of <becquerel.tools.isotope_qty.IsotopeQuantity object at 0x7f2bfb390410>> = <becquerel.tools.isotope_qty.IsotopeQuantity object at 0x7f2bfb390410>.atoms_now
tests/isotope_qty_test.py:271: AssertionError
```

[...]

PR #75, opened 2017-08-29
Fix issue 74

markbandstra:
Fixes issue #74 and adds some more isotopes to ```IsotopeQuantity``` tests.


tests/isotope_qty_test.py
```diff
@@ -497,7 +500,7 @@ def test_decay_normalize(radioisotope):
     interval1 = (now, now + datetime.timedelta(hours=1))
     assert np.isclose(decay_normalize(radioisotope, interval1, interval1), 1)
 
-    if radioisotope.half_life < 1000 * 3.156e7:     # see #65
+    if 90 < radioisotope.half_life < 1000 * 3.156e7:     # see #65
```
jccurtis: @markbandstra Can you add some comments for the `<` and `>` cases?

Issue #76, opened 2017-09-09
XCOM returns wrong data for request


jccurtis:
This unexpected test failure arose during [build 40.2](https://travis-ci.org/lbl-anp/becquerel/jobs/271779261) on python 3.6. The `TestFetchXCOMData` has a redundant `TestXCOMQuery` which queries XCOM for every test and one of the requests returned incorrect data:

```python
__________________________ TestFetchXCOMData.test_07 ___________________________
self = <xcom_test.TestFetchXCOMData object at 0x7f29fee2b240>
    def test_07(self):
        """Test fetch_xcom_data with compound (H2O) and three energies....."""
        energies = [60., 662., 1460.]
        xd = xcom.fetch_xcom_data('H2O', energies_kev=energies)
>       assert len(xd) == len(energies)
E       assert 100 == 3
E        +  where 100 = len(    energy  coherent  incoherent   photoelec  pair_nuc  pair_elec  \\n0      1.0   2.76100     0.00957  2295.00000     ...97       0.1974        0.1791  \n98       0.1955        0.1775  \n99       0.1936        0.1760  \n\n[100 rows x 8 columns])
E        +  and   3 = len([60.0, 662.0, 1460.0])
tests/xcom_test.py:224: AssertionError
``` 

I gave this test a shot and would be interested for your feedback @markbandstra:

```python
for i in range(100):
     df = bq.tools.xcom.fetch_xcom_data('H2O', energies_kev=[60., 662., 1460.])
     if len(df) != 3:
         print('=' * 100)
         print('*** TRIAL', i)
         print(df.shape)
         print(df)
```

The output gives different results across ranges of energies not passed to XCOM like:

```
*** TRIAL 42
(100, 8)
    energy  coherent  incoherent  photoelec  pair_nuc  pair_elec  total_w_coh  \
0    61.99   0.18410       3.175    0.10140       0.0        0.0        3.460
1    62.30   0.18240       3.173    0.09972       0.0        0.0        3.455
2    62.62   0.18070       3.171    0.09806       0.0        0.0        3.449
3    62.94   0.17900       3.169    0.09642       0.0        0.0        3.444
4    63.26   0.17730       3.167    0.09479       0.0        0.0        3.439
5    63.58   0.17560       3.164    0.09318       0.0        0.0        3.433
6    63.91   0.17390       3.162    0.09160       0.0        0.0        3.428
7    64.24   0.17230       3.160    0.09003       0.0        0.0        3.422
[...]

PR #78, opened 2017-09-12
Feature rebinning

### LINKED ISSUE #29: `Spectrum` rebinning
* Rebinning according to an input bin edges vector
* `rebin_like(spec)` to use the binning from another spectrum

Please expand on this, I'm not very familiar with rebinning methods or needs.

jccurtis:
Fixes #29

General:
- [x] @jccurtis write tests
  - [x] rebin_interpolation
  - [x] rebin_listmode
  - [x] edge cases (see below edges cases from @markbandstra)
  - [x] add tests to prevent bug (previously fixed)
- [x] @chunhochow better docstrings
- [x] @jccurtis Integrate into `Spectrum`
  - [x] `Spectrum.rebin`
  - [x] add tests
- [x] How to handle uncertainty propagation?
  - `cps` input always breaks
  - warn if using interpolation method
  - Raise `NotImplementedError`
- [x] @jccurtis Check types for numba
- [x] @cosama Add examples
- [x] @cosama Allow negative input/output channels

Moved to issues:

- [#90] Binomial sampling option (keyword argument)
- [#89] auto fit linear spline (keyword arg)
- [#88] auto fit higher-order spline? (with a `Sav-Gol` filter?)
- [#147] Allow spectrum.rebin with `force` keyword if the `counts` are not defined
- [#148] Add underflow/overflow include flag


chunhochow:
Travis CI is failing on a XCOM test...?!?

jccurtis:
@chunhochow NOPE, this is related to #76

jccurtis:
Testing needs to be added back. Somehow it was lost along the way ...

jccurtis:
I've added and updated the tests and fixed a reshaping bug. More tests need to be added:
- [ ] the `listmode` method
- [ ] edge cases for bin structures which don't completely overlap
- [ ] others?

becquerel/core/rebin.py
```diff
@@ -64,33 +64,40 @@ def _linear_offset(slope, cts, low, high):
...
+    Returns:
+      The indefinite integral of y = mx + b, with an x value substituted in.
+      (m x^2 / 2 + b x)
     '''
     return m * x**2 / 2 + b * x
 
 
 @nb.jit(nb.f8(nb.f8, nb.f8, nb.f8, nb.f8, nb.f8), nopython=True)
-def _counts(slope, offset, cts, low, high):
+def _counts(m, b, x_low, x_high):
```
jccurtis: @chunhochow travis failed because of the incorrect number of args in `_counts` compared to the type def in the `jit` decorator. Can you update?

markbandstra:
I would favor throwing a warning when rebinning into bins that are outside the original binned space. I can see this being a legitimate use, so an exception would be overkill, but it could potentially get someone into trouble if they didn't know.

I think an appropriate case to raise an exception would be if the new bins do not overlap the previous bins at all. This doesn't seem to be checked for right now.

In other words, warnings for:
```
old:   └┴┴┴┴┘
new: └┴┴┴┴┴┴┴┴┘
```
```
old:   └┴┴┴┴┘
new: └┴┴┴┘
```
```
old:   └┴┴┴┴┘
new:     └┴┴┴┴┘
```
and an exception for:
```
old:   └┴┴┴┴┘
new:          └┴┴┴┘
```

jccurtis:
Thanks @markbandstra! What about:

```
old:  └┴┴┴┴┴┴┴┴┘
new:    └┴┴┴┴┘
```

In this case the extra bins outside of the new space would be lost (we should check this in the tests). This boils down to the `min` and `max` edges not aligning... I think a warning could be thrown here as well. Thoughts?

cosama:
A bit off topic, but I just looked through the code and as far as I can see we have not had any warnings being displayed yet. How are we going to deal with warnings? Just a print or using the warning module?

At least to my opinion, we should at least make sure warnings are passed to stderr and not stdout.

jccurtis:
I think we should use the `warnings` module so the user is only notified once for an interpreter session instead of printing tons of times for a rebinning in a loop.

markbandstra:
I second using the [`warnings`](https://docs.python.org/3.1/library/warnings.html) module.

markbandstra:
Re: @jccurtis question about warning for
```
old:  └┴┴┴┴┴┴┴┴┘
new:    └┴┴┴┴┘
```

I think this is debatable. In this case you are throwing away data, but I think that is less bad than creating data where none exist (i.e., zero bins where there is no overlap). I could be convinced either way, but my initial thought is no warning.

cosama:
I agree to using the `warnings` module. However, I don't know how I feel about displaying histogram overlap messages just once. At least this is something that I want to know about every time it happens (or not at all, if I don't care). The first warning could mask a later warning that actually causes problems.

jccurtis:
@cosama, the `warnings` module gives user control over the printing of these warnings via filters, see: https://docs.python.org/3.6/library/warnings.html#the-warnings-filter

jccurtis:
@chunhochow 

- I fixed the bug in `rebin` for cases in which the `out_edges=in_edges[1:]` and the main loop exited without added any counts to the `out_spectrum` (see e88ef32)
- I updated from `develop` and resolved conflicts
- I added some more tests

We are getting closer. See above todo's

chunhochow:
@jccurtis wrt ` for in_idx in range(in_idx, len(in_spectrum)):` in `_rebin_interpolation`, I think it's on purpose, because the comment says `# loop through input bins (starting from the above in_idx value):`. It means that it moves `in_idx` on from the last value onward to `len(in_spectrum)`

chunhochow:
sure, if it doesn't complicate the logic too much. I don't remember how the logic runs so I'm not sure how big of a change that would be

jccurtis:
@cosama nice one!

markbandstra:
Weird, travis is giving the error `Branch "feature-rebinning" not included per configuration`, and it gave the same error for the previous two. Their [documentation](https://docs.travis-ci.com/user/pull-requests/#my-pull-request-isnt-being-built) says:

> So when you create or update a pull request, and Travis CI doesn’t create a build for it, make sure the pull request is mergeable. If it isn’t, rebase it against the upstream branch and resolve any merge conflicts. When you push the fixes up to the pull request, Travis CI will happily build them.

Can someone resolve the merge conflict with `becquerel/__init__.py` and push it?

jccurtis:
Building now after @cosama's merge resolution

cosama:
It seems that brought travis back to life.

jccurtis:
@cosama Change the rebin example notebook to the standard naming convention

jccurtis:
The testing is messy but could be ready to go. I'll check the notebook shortly. I added a flag (enabled by default) for the edge overlap warnings

cosama:
I think we can adjust the name of the notebook in #137, so if @jccurtis thinks this is good to go, we can merge it into `develop`.

becquerel/core/rebin.py
```diff
@@ -0,0 +1,433 @@
...
+    _check_ndim(in_spectra, {1, 2}, 'in_spectra')
+    # broadcast in_edges out to the dimensions of in_spectra
+    # specifically for the case 1D -> 2D
+    if (in_spectra.ndim == 2) and (in_edges.ndim == 1):
+        # copy req'd: the readonly array doesn't work w/ numba
+        in_edges = np.copy(
+            np.broadcast_to(
+                in_edges,
+                (in_spectra.shape[0], in_edges.shape[-1])))
+    _check_ndim(out_edges, 1, 'out_edges')
```
cosama: @jccurtis Well done putting this in, very elegant. Is there a particular reason you added a flag for `zero_pad`, wouldn't it be better to add a `no_checks` flag instead which removes all the checks in one go (for performance if necessary).

becquerel/core/rebin.py
```diff
@@ -0,0 +1,433 @@
...
+                np.int64, casting="safe", copy=False)
+        except TypeError:
+            raise ValueError(
+                "in_spectrum can only contain ints for method listmode")
+    else:
+        in_spectra = np.asarray(in_spectra, np.float)
+    in_edges = np.asarray(in_edges, np.float)
+    out_edges = np.asarray(out_edges, np.float)
+    # Check inputs
+    # TODO check that in_spectrum are all >= 0
```
markbandstra: Implement this check only for `method='listmode'`

becquerel/core/rebin.py
cosama: Flag suppresses warnings, other tests raise errors which are not disabled. That makes sense.

becquerel/core/spectrum.py
```diff
@@ -784,6 +786,58 @@ def combine_bins(self, f):
...
+        zero_pad_warnings (boolean): warn when edge overlap results in
+            appending empty bins
+
+        Raises:
+            SpectrumError: for bad input arguments
+
+        Returns:
+            A new Spectrum object with the rebinned data.
+        """
+        # TODO @jccurtis handle this differently?
```
markbandstra: @jccurtis clean up

becquerel/core/spectrum.py
```diff
@@ -784,6 +786,58 @@ def combine_bins(self, f):
...
+        if self.bin_edges_kev is None:
+            raise SpectrumError('Cannot rebin spectrum without energy '
+                                'calibration')
+        else:
+            in_edges = self.bin_edges_kev
+        if self.counts is None:
+            # NOTE @jccurtis this will be deprecated when closing #122
```
markbandstra: @jccurtis clean up please

becquerel/core/spectrum.py
```diff
@@ -784,6 +786,58 @@ def combine_bins(self, f):
...
+            raise SpectrumError('Cannot rebin method without `counts` data')
+        if method.lower() == 'interpolation':
+            in_spec = self.counts_vals.astype(np.float)
+        elif method.lower() == 'listmode':
+            in_spec = self.counts_vals.astype(np.int)
```
markbandstra: Type casting is now done inside `rebin` function.

becquerel/core/spectrum.py
```diff
@@ -784,6 +786,58 @@ def combine_bins(self, f):
...
+        else:
+            raise SpectrumError('Unknown rebining method: {}'.format(method))
```
markbandstra: Spelling :)

becquerel/core/rebin.py
```diff
@@ -0,0 +1,433 @@
...
+    for i in np.arange(in_spectra.shape[0]):
+        out_spectra[i, :] = _rebin_listmode(in_spectra[i, :], in_edges[i, :],
+                                            out_edges, slopes[i, :])
+    return out_spectra
+
+
+def rebin(in_spectra, in_edges, out_edges, method="interpolation",
+          slopes=None, zero_pad_warnings=True):
+    """
+    Spectra rebining via deterministic or stochastic methods.
```
markbandstra: Spelling :)

becquerel/core/rebin.py
```diff
@@ -0,0 +1,433 @@
...
+    if in_spectra.ndim == 2:
+        # Check number of spectra
+        assert in_spectra.shape[0] == in_edges.shape[0], (
+            "Number of in_spectra({}) differs from number of in_edges({})"
+        ).format(in_spectra.shape, in_edges.shape)
+    # Update dtypes if necessary
+    in_edges = np.asarray(in_edges, dtype=np.float64)
+    out_edges = np.asarray(out_edges, dtype=np.float64)
+    slopes = np.asarray(slopes, dtype=np.float64)
+    # Specific calls to different rebining methods
```
markbandstra: Spelling :)

becquerel/core/spectrum.py
```diff
@@ -784,6 +786,58 @@ def combine_bins(self, f):
         obj = Spectrum(**kwargs)
         return obj
 
+    def rebin(self, out_edges, method="interpolation", slopes=None,
+              zero_pad_warnings=True):
+        """
+        Spectra rebining via deterministic or stochastic methods.
```
markbandstra: Spelling :)

jccurtis:
@markbandstra @cosama This is ready to go!

becquerel/core/spectrum.py
```diff
@@ -806,6 +807,55 @@ def combine_bins(self, f):
...
+
+        Returns:
+            A new Spectrum object with the rebinned data.
+        """
+        if self.bin_edges_kev is None:
+            raise SpectrumError('Cannot rebin spectrum without energy '
+                                'calibration')
+        if ((method.lower() == 'listmode') and
+                (self._counts is None) and
+                (self.livetime is not None)):
```
cosama: Not sure if the last check is needed, maybe better to leave it out and have the warning here. If livetime is not defined it will fail further down (at `self.counts_vals`) anyways and the warning helps to see what is going on here.

becquerel/core/rebin.py
```diff
@@ -0,0 +1,436 @@
...
+            warnings.warn(
+                'Converting in_spectra to integers for rebin method listmode',
+                RebinWarning)
+        in_spectra = np.asarray(in_spectra, dtype=np.int64)
+    else:
+        in_spectra = np.asarray(in_spectra, dtype=np.float)
+    in_edges = np.asarray(in_edges, np.float)
+    out_edges = np.asarray(out_edges, np.float)
+    # Check inputs
+    if method == 'listmode':
```
cosama: Move to previous `if method == 'listmode'` statement.

becquerel/core/rebin.py
```diff
@@ -0,0 +1,436 @@
...
+    if method == "interpolation":
+        in_spectra = np.asarray(in_spectra, dtype=np.float64)
+        if in_spectra.ndim == 2:
+            return _rebin2d_interpolation(
+                in_spectra, in_edges, out_edges, slopes)
+        elif in_spectra.ndim == 1:
+            return _rebin_interpolation(
+                in_spectra, in_edges, out_edges, slopes)
+    elif method == "listmode":
+        in_spectra = np.asarray(in_spectra, dtype=np.int64)
```
cosama: Didn't we do the conversion earlier?

jccurtis:
Okay, resolved @cosama's comments, cleaned up legacy assertion errors and simplified/grouped checks

jccurtis:
Ready for merging!

becquerel/core/spectrum.py
```diff
@@ -806,6 +807,57 @@ def combine_bins(self, f):
...
+                                'calibration')
+        in_spec = self.counts_vals
+        if method.lower() == 'listmode':
+            if (self._counts is None) and (self.livetime is not None):
+                warnings.warn(
+                    'Rebinning by listmode method without explicit counts ' +
+                    'provided in Spectrum object',
+                    SpectrumWarning)
+            # NOTE: @jccurtis asks: why are the counts not integers already?
+            in_spec = in_spec.astype(np.int64)
```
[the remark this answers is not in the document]
jccurtis: I added this intentionally because a warning is raised since the `Spectrum.count_vals` are `float` values. My note was a general question about why `Spectrum.count_vals` are `float`'s. You do bring up the obvious reason why this would be the case: `interpolation` rebinning ...

So we need to decide if `listmode` rebinning with `float` spectrum input raises a warning ...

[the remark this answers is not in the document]
jccurtis: Okay one idea is to check the decimal residual on the float inputs:

```python
if np.any(np.abs(counts - np.fix(counts)) > 1e-6):
    raise RebinError()
```

along the same lines, if all values are less than one, raise an error:

```python
if np.all(counts < 1):
    raise RebinError()
```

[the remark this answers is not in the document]
cosama: Isn't that what the check on line 399 in rebin.py is for?

I kind of like your idea about checking if the float values could originate from a integer array. However, it probably would make more sense in rebin.py or is this something that is specific to Spectrum somehow? Also I'm not sure how numpy handles float -> integer conversion, but it probably would be safer if we do np.fix(counts+0.5). Also could we use a np.allclose(counts - np.fix(counts+0.5)) or is there something fishy about that?

PR #102, opened 2018-04-10
Fix #86

### LINKED ISSUE #86: Invalid energy coeffs prevent loading file
If an SPE file has the following tag:

```
$MCA_CAL:
3
0.000000E+000 0.000000E+000 0.000000E+000 
```

A `ValueError` is raised upon loading that will prevent usage of the data file. I propose that we change this `error` to a `warning` and don't set the energy calibration for the output `Spectrum` object. 

[Example spe file (changed ext to txt to upload)](https://github.com/lbl-anp/becquerel/files/1712788/digibase_5min_30_1.txt)

_Edited to correct spe fieldname_

markbandstra:
Fixes #86 by removing an invalid calibration read from file.


markbandstra:
Looks like I forgot to include the sample file to actually test this bug.

PR #108, opened 2018-04-10
Fix #96 (NNDC kwargs conflict)

### LINKED ISSUE #96: fetch_wallet_card kwarg overwrite
@markbandstra and I just found a bug with to `nndc.fetch_wallet_card` when specifying `z` and `a_range`. It looks like the `z` kwarg is overwritten

markbandstra:
Minor update to NNDC query API that is transparent to the user, plus added more checks to the NNDC tests. Affects both `fetch_wallet_card` and `fetch_decay_radiation`.

Fixes #96

### DIFF
--- a/becquerel/tools/nndc.py
+++ b/becquerel/tools/nndc.py
@@ -425,8 +425,9 @@ class _NNDCQuery(object):
         for x in ['z', 'a', 'n']:
             # handle Z, A, and N settings
             if x in kwargs:
-                self._data['spnuc'] = 'zan'
-                self._data[x.lower()] = '{}'.format(kwargs[x])
+                self._data['spnuc'] = 'zanrange'
+                self._data[x + 'min'], self._data[x + 'max'] = \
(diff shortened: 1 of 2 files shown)


jccurtis:
@markbandstra, so in the case of a single query a `DataFrame` is still returned instead of a `Series`?

markbandstra:
Yes, any query always returns a `DataFrame`. Not sure how a `Series` would make sense for mixed data types.

jccurtis:
Thank you! I just wanted to confirm that a `DataFrame` is always returned. What do you think about moving `ufloats_overlap_range` to a general test location since this will likely be useful for other tests?

jccurtis:
Maybe the test helper functions could be located inside `bq.testing`

markbandstra:
We could reorganize `ufloats_overlap_range` but I am not sure if that is needed now. I wrote this function so that a couple of the NNDc tests would pass after I started checking the limits.

PR #109, opened 2018-04-10
Update gitignore from github/gitignore

markbandstra:
Pytest recently started creating hidden files that are not covered by the current gitignore, so I updated the whole file using github/gitignore/Python.gitignore


jccurtis:
@markbandstra I merged develop in and will merge after travis passes

Issue #37, opened 2017-03-23
Spectrum plotting


bplimley:
Methods in `Spectrum` for plotting. Visualization is not a main focus, but at least basic plotting with some common options.

Some ideas:
* use `drawstyle='steps-mid'` for "stairs"-type plot
* log y-scale by default
* automatically label axes
* color, legend label, title
* allow plotting into an existing axes

cosama:
I just submitted the newest version, it includes errorbars/bands too. Have a look, if it is good I can send out the pullrequest.

cosama:
Merging pull request #114 closes this issue.

PR #114, opened 2018-05-24
Feature plotting #37

cosama:
This closes issue #37.

It adds a SpectrumPlotter class, that allows for plotting a Spectrum with matplotlib's plot and fill_between methods.  It has its own way of creating a step like function, which should handle edges correctly for Becquerel style Spectra. It deals with labels and titles automatically if not otherwise requested. The user can provide limits and yscale if he wants to alter the behavior of the class. It also  has functions to plot errorbands and errorbars. For some example see the respective [notebook](https://github.com/lbl-anp/becquerel/blob/a78e7205f72f885b8ee60341db6deabea129e313/examples/plotting_examples.ipynb).

It also adds a plot and fill_between method to Spectrum which uses the SpectrumPlotter class to create plots.

I added tests for most of the stuff, that do not require plots to pop up and thus can be run in travis. It might be a good idea to check on the tests and see if they are sufficient.

Let me know if it need any changes.


cosama:
I just had a look at it, it has to do with matplotlibs backend. I just commited a version that should fix it (it shows up as Chun Ho because I did it on the Astro)

cosama:
No it did not work. It builds locally, but it seems not to like the travis backend of matplotlib. I will see if there is a fix, otherwise we might be forced to turn off all the tests.

cosama:
The travis build works now for both python 2 and python 3. Before approving, please have a look at the new counts_over methods in Spectrum. I am really not sure if they add any useful functionality, also I just saw some of these method need some fixing up (proper descriptions).

jccurtis:
As @cosama points out there was some additional `Spectrum` methods added in baa72f6551b0c21e557cb8c86b576255723ba7ef and used to plot in 24babbebc60fc5b60b3b505155f86724eede5a66. I think we should moved these time scaling parameters to another branch and think about how to better use this scaling (or abandon it).

cosama:
I moved the changes to the Spectrum class (the new methods to evaluate a cps spectrum over a certain time) to a new branch (eval-over-hack) and removed it from this branch.

I'm certain there are a few bugs left, but I think overall it would be ready to be merged to develop.

Issue #117, opened 2018-06-29
EnergyCal should perform fit on initialization or report proper Errors


cosama:
If one creates an energy calibration object at this point and does not call the update_fit function one gets an obscure KeyError about coeffs when applying the calibration to a spectrum, that does not help at all to understand what is going on (at least I was very confused).

Either we catch the error and tell the users that they need to call the update_fit function for it to work, or we call the update_fit function on initialization (my prefered choice) and make sure the fit coefficients are always defined properly.

jccurtis:
Can you upload some sample code? I agree that this should be in the `from_points` constructor but **not** in the base init because you can make a calibration from coefficients and not the calibration points. What do you think? 

Also, this test could be updated: https://github.com/lbl-anp/becquerel/blob/develop/tests/energycal_test.py#L215-L219

cosama:
The test you found it exactly what I was referring too. I agree it should not be in the init, but the from_points method should call it automatically. That also won't break backwards compatibility (if we even care) as you still can manually call the `update_fit` function later, without affecting the result.

cosama:
I just realized that from_points does not automatically include (0,0). Is that something we want, otherwise we should add that to the docu string (my preference).

jccurtis:
Let’s add add a keyword for `zero_crossing` or `include_origin` to `from_points`. I could make a quick PR after lunch.

cosama:
Already working on it. I can do that.

jccurtis:
Excellent. Thank you!

PR #118, opened 2018-06-29
Bugfix energycal

cosama:
Fixes issue #117.

`from_point` does execute the function `update_fit()` internally. Furthermore, there is a flag to include the origin/(0,0) into the fit, without adding it to the list. I also added a manual linear regression to the tests, to check if the coefficients are calculated correctly.

### DIFF
--- a/becquerel/core/energycal.py
+++ b/becquerel/core/energycal.py
@@ -49,8 +49,6 @@ class EnergyCalBase(object):
     def from_points(cls, chlist, kevlist, include_origin=False):
         """Construct EnergyCal from calibration points.
 
-        Does not include a point at zero automatically.
-
         Args:
           chlist: list/tuple/array of the channel values of calibration points
           kevlist: list/tuple/array of the corresponding energy values [keV]
--- a/becquerel/core/energycal.py
+++ b/becquerel/core/energycal.py
@@ -61,7 +61,7 @@ class EnergyCalBase(object):
 
         try:
             cond = len(chlist) != len(kevlist)
-        except:
+        except TypeError:
             raise BadInput('Inputs must be one dimensional iterables')
         if cond:
             raise BadInput('Channels and energies must be same length')
--- a/becquerel/core/energycal.py
+++ b/becquerel/core/energycal.py
@@ -46,12 +46,16 @@ class EnergyCalBase(object):
         # initialize fit constraints?
 
     @classmethod
-    def from_points(cls, chlist, kevlist):
+    def from_points(cls, chlist, kevlist, include_origin=False):
         """Construct EnergyCal from calibration points.
+
+        Does not include a point at zero automatically.
 
         Args:
           chlist: list/tuple/array of the channel values of calibration points
           kevlist: list/tuple/array of the corresponding energy values [keV]
+          include_origin: Default is False, if set to True will add a point
+                          at zero into the fit.
 
         Raises:
           BadInput: for bad chlist and/or kevlist.
@@ -59,8 +63,20 @@ class EnergyCalBase(object):
 
         if chlist is None or kevlist is None:
             raise BadInput('Channel list and energy list are required')
-        elif (not isinstance(chlist, VECTOR_TYPES) or
-                not isinstance(kevlist, VECTOR_TYPES)):
+
+        cond1 = isinstance(chlist, VECTOR_TYPES)
+        cond2 = isinstance(kevlist, VECTOR_TYPES)
+
+        if include_origin:
+            if cond1 and cond2:
+                chlist = np.append(0, chlist)
+                kevlist = np.append(0, kevlist)
+            elif not cond1 and not cond2:
+                chlist = [0, chlist]
+                kevlist = [0, kevlist]
+                cond1 = cond2 = True
+
+        if (not cond1 or not cond2):
             raise BadInput('Inputs should be vector iterables, not scalars')
         elif len(chlist) != len(kevlist):
             raise BadInput('Channels and energies must be same length')
@@ -68,6 +84,7 @@ class EnergyCalBase(object):
         cal = cls()
         for ch, kev in zip(chlist, kevlist):
             cal.new_calpoint(ch, kev)
+        cal.update_fit()
         return cal
 
     @classmethod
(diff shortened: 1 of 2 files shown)


cosama:
It didn't build in python2. Let me have a look at that and fix it. I think it might still make sense to catch the KeyError and give a more meaningful message if one tries to calibrate a spectrum that has not all parameters set. Let me see if I can add that and write another test for it.

tests/energycal_test.py
```diff
@@ -69,6 +69,18 @@ def cal_spec(spec_data):
 
     return bq.Spectrum(spec_data, bin_edges_kev=TEST_EDGES_KEV)
 
+@pytest.fixture
```
jccurtis: @cosama I don't think you want function to be a fixture. This is just a support method right? Also, per @markbandstra's comment regarding #78, how about including some fixed input/output (i.e. the gain should be `1` for points (1,1), (2,2)).

cosama: Thanks for the comment. I added a test for chlist (0,1) at kevlist (1,3) now (to also get a non-zero offset) and removed the fixture.

becquerel/core/energycal.py
```diff
@@ -46,28 +46,43 @@ def __init__(self):
...
+        cond1 = isinstance(chlist, VECTOR_TYPES)
+        cond2 = isinstance(kevlist, VECTOR_TYPES)
+
+        if include_origin:
+            if cond1 and cond2:
+                chlist = np.append(0, chlist)
+                kevlist = np.append(0, kevlist)
+            elif not cond1 and not cond2:
+                chlist = [0, chlist]
+                kevlist = [0, kevlist]
```
jccurtis: @cosama, I think this will allow scalar `chlist` and `kevlist` which is confusing because that will `raise` a `BadInput` exception below if `inlucude_origin=False`. Did you have a specific case in mind when including this `elif`? Otherwise I think this logic could be simplified to only allow `VECTOR_TYPES` input

becquerel/core/energycal.py
```diff
@@ -332,13 +347,21 @@ def valid_coeffs(self):
     def slope(self):
         """Return the slope coefficient value."""
 
-        return self._coeffs['b']
+        try:
+            return self._coeffs['b']
+        except KeyError:
```
jccurtis: This could also be done with:

```python
if 'b' in self._coeffs:
    return self._coeffs['b']
else:
    raise EnergyCalError('Slope coefficient not yet supplied or calculated.')
```

But I think the `KeyError` catch is more elegant

chunhochow: i agree with using `KeyError`

jccurtis:
Besides the `from_points` input I think this PR is ready

becquerel/core/energycal.py
```diff
@@ -46,28 +46,43 @@ def __init__(self):
         # initialize fit constraints?
 
     @classmethod
-    def from_points(cls, chlist, kevlist):
+    def from_points(cls, chlist, kevlist, include_origin=False):
```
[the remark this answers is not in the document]
markbandstra: Consensus here after discussion is to disallow scalar inputs (and add a test that scalar inputs raise BadInput), and keep `include_origin`.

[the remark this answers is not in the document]
markbandstra: Other BadInputs: multidimensional numpy arrays (test this).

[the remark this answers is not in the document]
jccurtis: Thanks all, what would be the point of multi dimensional inputs?

tests/energycal_test.py
```diff
@@ -70,6 +70,18 @@ def cal_spec(spec_data):
     return bq.Spectrum(spec_data, bin_edges_kev=TEST_EDGES_KEV)
 
 
+def linear_regression(x, y):
```
markbandstra: Why not use [`numpy.polyfit(x, y, 1)`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.polyfit.html)?

cosama: is used internally, I thought it would add some more confidence if another method produces the same result, but by now I tend to leave these tests out anyways, as it is covert sort of in 

 ```
def test_linear_fitting_simple():
    """Test fitting of known slope and offset"""

    cal = bq.LinearEnergyCal.from_points(chlist=[0.0, 1.0], kevlist=[1.0, 3.0])
    assert np.allclose([1.0, 2.0], [cal.offset, cal.slope])
```

**Comment by cosama** [comment]
While it covers most of the wrong inputs I can think of right now, I do not particular like the solution with np.append() as it is now. However, I can also not come up with a better way that still covers all those cases efficiently. Any opinions on that?

**Comment by cosama** [comment]
I did quite a bit of research on the input issues and I came to the conclusion, that we should only check input if there is no other way (as for example to make sure both items in a zip() are or the same length). The proper way to deal with anything else, is to catch the errors that is caused further along in the code, if the message the it causes is not clear by itself. This is the most pythonic way and in this case also much clearer than the append method. I changed that now, have a look. If you agree with me on this matter we might want to add that to the style guide or something.

File: becquerel/core/energycal.py [hunk 6]
```diff
@@ -60,11 +60,10 @@ def from_points(cls, chlist, kevlist, include_origin=False):
             raise BadInput('Channel list and energy list are required')
 
         try:
-            len1 = len(chlist)
-            len2 = len(kevlist)
+            cond = len(chlist) != len(kevlist)
         except:
```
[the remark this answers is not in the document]
cosama: Now I feel silly. Good catch.

tests/energycal_test.py
```diff
@@ -276,3 +346,10 @@ def test_rm_calibration_error(uncal_spec):
     assert not uncal_spec.is_calibrated
     uncal_spec.rm_calibration()
     assert not uncal_spec.is_calibrated
+
+def test_callibration_not_inizialized_error(uncal_spec):
```
markbandstra: Please check spelling :)

cosama: Let me know if you find another one. Changed  a few docu strings in the tests.

tests/energycal_test.py
```diff
@@ -276,3 +346,10 @@ def test_rm_calibration_error(uncal_spec):
     assert not uncal_spec.is_calibrated
     uncal_spec.rm_calibration()
     assert not uncal_spec.is_calibrated
+
+def test_callibration_not_initialized_error(uncal_spec):
```
markbandstra: calibration 😬

Issue #121, opened 2018-07-03
Add CONTRIBUTING.md


markbandstra:
Many repos I have seen have guides for contributions, e.g.:

https://github.com/numpy/numpy/blob/master/CONTRIBUTING.md
https://github.com/numba/numba/blob/master/CONTRIBUTING.md
https://github.com/github/gitignore/blob/master/CONTRIBUTING.md

I think we might want to move the Code Style Guide and Linter sections out of our README and into such a file. It might also be helpful to add a checklist to describe the workflow, e.g.:

- [ ] Branch off of `develop` and name the branch `feature-XX` or `issue-XX`
- [ ] Develop your feature or fix
- [ ] Write tests to cover all use cases
- [ ] Ensure test coverage is >95%
- [ ] Ensure all tests pass (`python setup.py test`)
- [ ] Ensure there are no linter issues and style is consistent
- [ ] Push branch to GitHub
- [ ] Create a pull request

jccurtis:
Additionally we could add linting to the CI build which will reduce manual checking of style enforcement, check [this](https://jeffknupp.com/blog/2016/12/09/how-python-linters-will-save-your-large-python-project/) out.

markbandstra:
Good point, I would be in favor of that. We should probably modify our pylintrc to remove the most annoying and unnecessary warnings though.

jccurtis:
Definitely, lets chat about how that will look next Tuesday as well. 😉

Issue #119, opened 2018-07-03
Upload becquerel to PyPI


markbandstra:
We should get this package on the [Python Package Index](https://pypi.org) so that anyone can `pip install` it. There are steps [here](https://packaging.python.org/tutorials/packaging-projects/) for this process.

It looks like we need to do some work on [`setup.py`](https://github.com/lbl-anp/becquerel/blob/61e6b2d0b3c4eba134425e7c17c3fb51990f06c6/setup.py) before we do this:
  - [x] update version number
  - [x] add `maintainer`/`maintainer_email` (this can be me)
  - [x] update `long_description` to read from the README
  - [x] add `long_description_content_type="text/markdown"`
  - [x] fix `url` link (no longer works)
  - [x] fix `download_url` link (no longer works)
  - [x] update `classifiers`
  - [x] add `license` (should this be the text of LICENSE.txt?)
  - [x] try `packages=setuptools.find_packages()`
  - [x] add any other relevant [`setup()` arguments](https://packaging.python.org/guides/distributing-packages-using-setuptools/#setup-args)

In addition, we should add a description to the top of the README of what `becquerel` is and what is its purpose. (Right now it just starts with installation instructions.)

Should we coordinate this with a new tagged release, say version 0.2?

jccurtis:
Thanks for putting this together @markbandstra! I agree that this should go along in v0.2.0 along with the fitting module.

cosama:
Yeah, great work getting that started. I think it is a good idea to make the software easier accessible, but I think we should try to first get at least rebinning (#78), maybe also autocalib (#118) working. I consider those two to be essential for some basic spectroscopy user cases.

chunhochow:
If we move to v1.0 we might trigger the IPO process.

jccurtis:
FYI you can install with:

```bash
pip install git+https://github.com/lbl-anp/becquerel
```

PR #124, opened 2018-07-11
Updates to package for release

markbandstra:
* Addresses #119 and #121
* Updates and adds additional package metadata
* Uses `setup.py` to manage some metadata to reduce repetition
* Establishes checklists for contributing code and releasing new versions
* Improves gitignore for multiple platforms

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,23 @@
+## Contribution guidelines
+
+Contributions to `becquerel` are welcome and encouraged, whether it is
+reporting bugs, requesting features, or contributing code.
+Please follow these guidelines when contributing to this project.
+
+### Reporting issues
+
+When reporting issues with `becquerel`, please provide a minimum working example to help identify the problem and tag the issue as a `bug`.
+
+### Feature requests
+
+For a feature request, please create an issue and label it as a `new feature`.
+
+### Checklist for code contributions:
+  - [ ] Branch off of `develop` and name the branch `feature-XX` or `issue-XX`
+  - [ ] Develop the feature or fix
+  - [ ] Write tests to cover all use cases
+  - [ ] Ensure test coverage is >95%
+  - [ ] Ensure all tests pass (`python setup.py test`)
+  - [ ] Ensure there are no linter issues (both `flake8` and `pylint`)
+  - [ ] Check style is [consistent with Google Python style guide](https://google.github.io/styleguide/pyguide.html)
+  - [ ] Push branch to GitHub and create a pull request
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,0 +1,13 @@
+## Release history
+
+### Version 0.2 (2018-XX-XX)
+
+#### Improvements
+
+* Arma virumque cano
+* Troiae qui primus ab oris
+
+#### Bug fixes
+
+* Italiam fato profugus
+* Laviniaque venit
--- a/README.md
+++ b/README.md
@@ -1,5 +1,18 @@
 # becquerel
 
+## Overview
+
+Becquerel is a Python package for analyzing nuclear spectroscopic
+measurements. The core functionalities are reading and writing different
+spectrum file types, fitting spectral features, performing detector
+calibrations, and interpreting measurement results. It includes tools for
+plotting radiation spectra as well as convenient access to tabulated nuclear
+data, and it will include fits of different spectral features. It relies
+heavily on the standard scientific Python stack of numpy, scipy, matplotlib,
+and pandas. It is intended to be general-purpose enough that it can be useful
+to anyone from an undergraduate taking a laboratory course to the advanced
+researcher.
+
 ## Installation instructions
 
 ### As a user
--- a/README.md
+++ b/README.md
@@ -65,15 +65,6 @@ with line-by-line highlighting:
 pytest --cov-report html:htmlcov
 ```
 
-## Code Style Guide
-
-Use [google standards](https://google.github.io/styleguide/pyguide.html)
-
-## Linter
-
-* Use `flake8` in your IDE
-* Use `pylint` from command line (as in style guide)
-
 ## Copyright Notice
 Becquerel v. 0.1, Copyright (c) 2017, The Regents of the University of California (UC), through Lawrence Berkeley National Laboratory, and the UC Berkeley campus (subject to receipt of any required approvals from the U.S. Dept. of Energy). All rights reserved.
 If you have questions about your rights to use or distribute this software, please contact Berkeley Lab's Innovation & Partnerships Office at  IPO@lbl.gov.
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,12 @@
+## Release checklist
+
+- [ ] Branch off of `develop` and name the branch `version-X.X.X`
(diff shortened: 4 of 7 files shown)


jccurtis:
Thanks for putting this together @markbandstra! It looks like the py36 built failed on accessing XCOM. 🤦‍♂️ This is probably just a connection error. Can you run the tests locally?

markbandstra:
Weird, other tests are failing, and none of these fail for me locally.

CONTRIBUTING.md
```diff
@@ -0,0 +1,23 @@
...
+
+### Reporting issues
+
+When reporting issues with `becquerel`, please provide a minimum working example to help identify the problem and tag the issue as a `bug`.
+
+### Feature requests
+
+For a feature request, please create an issue and label it as a `new feature`.
+
+### Checklist for code contributions:
```
cosama: Maybe we should add ` [  ] Spellcheck your code and docu-strings`. I know at least one person not doing so :).

README.md
```diff
@@ -1,5 +1,16 @@
 # becquerel
```
cosama: Maybe we should also point out that to install becquerel it is best to use the `--user` flag, which puts it into a user owned directory, compared to the way described below (which often requires root access). It might even make sense to add this as the default way of installing it (not sure if it is cross platform compatible). At least for newcomers that might be very valuable information.

RELEASING.md
```diff
@@ -0,0 +1,21 @@
+## Release checklist
+
+- [ ] Branch off of `develop` and name the branch `version-X.X.X`
+- [ ] Update version number within the repository
+  - in `setup.py`
+  - in Copyright Notice in `README`
+  - in the `LICENSE`
+- [ ] Update classifiers in `setup.py`
+- [ ] Update `HISTORY` with a list of changes for this version
+- [ ] Verify that all tests pass (`python setup.py test`)
```
cosama: While the code should follow style guide all the time, I think it should be checked (`pylint` and `flake8`) here in particular. Also spelling might make sense to be checked here again. Just to make sure we have a clean release.

RELEASING.md
```diff
@@ -0,0 +1,21 @@
+## Release checklist
```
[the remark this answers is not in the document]
markbandstra: I personally would prefer having two separate sets of instructions: one for "external" folks who want to add a feature or fix a bug, versus the steps the core developers should go through to tag a release. They seem conceptually different to me.

HISTORY.md
```diff
@@ -0,0 +1,13 @@
+## Release history
```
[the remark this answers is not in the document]
markbandstra: No, not necessary. It would take effort to maintain, and you are right that all of this information is in the git history and on GitHub.

CONTRIBUTING.md
markbandstra: Good point. Maybe also requiring `pydocstyle` to check the docstrings.

README.md
markbandstra: Good point.

RELEASING.md
markbandstra: Fair point. If these checks are done upstream (perhaps included in continuous integration) then it shouldn't need to be done here. But might be good to check again.

markbandstra: By the way, we need to do a fair amount of work just to get the current code up to snuff with `pylint`.

cosama: I moved this discussion to #125. Maybe just mention it in RELEASING might still be worth it, it is easy to cross off a box as long as there are not too many yet.

HISTORY.md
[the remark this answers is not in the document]
cosama: I would remove it, but we should definitely require these information being added to the tag (needs to be part of RELEASING).

CONTRIBUTING.md
```diff
@@ -0,0 +1,23 @@
...
+  - [ ] Branch off of `develop` and name the branch `feature-XX` or `issue-XX`
+  - [ ] Develop the feature or fix
+  - [ ] Write tests to cover all use cases
+  - [ ] Ensure test coverage is >95%
+  - [ ] Ensure all tests pass (`python setup.py test`)
+  - [ ] Ensure there are no linter issues (both `flake8` and `pylint`)
+  - [ ] Check style is [consistent with Google Python style guide](https://google.github.io/styleguide/pyguide.html)
```
cosama: Just realized that the style guide is deprecated, we probably should also move to https://github.com/google/styleguide/blob/gh-pages/pyguide.md

README.md
markbandstra: This change was made with 3f416cf (comment remains visible since another line was changed).

jccurtis:
@markbandstra there are some useful bits [here](https://github.com/pypa/sampleproject/blob/master/tox.ini) for extra tests for `pypi` compat

CONTRIBUTING.md
```diff
@@ -0,0 +1,24 @@
...
+### Reporting issues
+
+When reporting issues with `becquerel`, please provide a minimum working example to help identify the problem and tag the issue as a `bug`.
+
+### Feature requests
+
+For a feature request, please create an issue and label it as a `new feature`.
+
+### Checklist for code contributions:
+  - [ ] Branch off of `develop` and name the branch `feature-XX` or `issue-XX`
```
markbandstra: Does anyone have an opinion about branch naming conventions? I wrote this just as a proposal.

cosama: I think the way you phrased it is good. I just didn't adhere to it so far.

RELEASING.md
```diff
@@ -0,0 +1,21 @@
...
+
+- [ ] Branch off of `develop` and name the branch `version-X.X.X`
+- [ ] Update version number within the repository
+  - in `setup.py`
+  - in Copyright Notice in `README`
+  - in the `LICENSE`
+- [ ] Update classifiers in `setup.py`
+- [ ] Verify that all tests pass (`python setup.py test`)
+- [ ] Commit the changes, push to GitHub, and start a pull request
```
jccurtis: @markbandstra this PR should be into master

RELEASING.md
```diff
@@ -0,0 +1,21 @@
...
+- [ ] Add release notes to the tag on GitHub with a list of changes
+- [ ] Upload new version to PyPI
+  - `python3 -m pip install --user --upgrade setuptools wheel`
+  - `python3 setup.py sdist bdist_wheel`
+  - `python3 -m pip install --user --upgrade twine`
+  - `python3 -m twine upload dist/*`
+- [ ] Test new version installs from PyPI
+  - `cd ..`
+  - `python3 -m pip install becquerel`
+  - `python3 -c "import becquerel; print(becquerel.__version__)"`
```
jccurtis: @markbandstra some more steps at the end:
- [ ] create PR from `release` branch to `develop`
- [ ] delete `release` branch

For reference this is using the methodology found in [git flow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow)

RELEASING.md
```diff
@@ -0,0 +1,21 @@
+## Release checklist
+
+- [ ] Branch off of `develop` and name the branch `version-X.X.X`
```
markbandstra: Let's change this to `release-X.X.X`

Issue #125, opened 2018-07-11
Add linting to pytests


cosama:
As @jccurtis pointed out, we should think about adding linting to the pytests. The issues raised by `pylint` can be questioned for their validity, but there is definitely plenty to improve:

```
+-----------------------------+------------+
|message id                   |occurrences |
+=============================+============+
|len-as-condition             |18          |
+-----------------------------+------------+
|fixme                        |13          |
+-----------------------------+------------+
|line-too-long                |11          |
+-----------------------------+------------+
|no-else-return               |9           |
+-----------------------------+------------+
|too-many-instance-attributes |6           |
+-----------------------------+------------+
|too-many-statements          |5           |
+-----------------------------+------------+
|too-many-branches            |5           |
+-----------------------------+------------+
|redefined-builtin            |3           |
+-----------------------------+------------+
|duplicate-code               |3           |
+-----------------------------+------------+
|protected-access             |2           |
+-----------------------------+------------+
|abstract-method              |2           |
+-----------------------------+------------+
|wrong-import-order           |1           |
+-----------------------------+------------+
|unused-variable              |1           |
+-----------------------------+------------+
|too-many-public-methods      |1           |
+-----------------------------+------------+
|too-many-locals              |1           |
+-----------------------------+------------+
|too-many-arguments           |1           |
+-----------------------------+------------+
|superfluous-parens           |1           |
+-----------------------------+------------+
|redefined-outer-name         |1           |
+-----------------------------+------------+
|no-self-use                  |1           |
+-----------------------------+------------+

```

The top on the list items are easy to fix. Also I think we should be flexible with line length. 80 characters seems kind of short nowadays with these large screens everywhere. I think it is a good length to aim for, but we should probably be flexible and allow for a bit more, say 90-100. 

The `too-many` are hard to deal with and I think we should disable those, some of the limits are pretty random anyways. A spectrum or the plotting class just need lots of methods and instances, and often it is hard to limit branches and statements.

The others are probably worth dealing with on case-by-case basis, I haven't checked the code for all of those.

Also we should think about if we use `pylint` on tests-fles too. Some of the tests cause errors in `pylint`, not just warnings, which is definitely an issue if we want to do that. `pyline` seems to not understand all the `pytest` syntax. Is there a special `pylint` instant or something for `pytest`?

jccurtis:
[...]

PR #126, opened 2018-07-12
Make code pylint compatible and integrate pylint into tests

cosama:
Fixes issue #125.

I started working on the core modules, which should be good now. Have a look if this is pleasing and is a style that we can all agree to.

Before it can be pulled in we have to:
- [x] Fix core module
- [ ] Fix parser module
- [ ] Fix tools module
- [ ] Deal with fixme's and TODO's
- [ ] Figure out how to check tests
- [ ] Pull in most recent develop
- [ ] Add pylint support to tests

### DIFF
--- a/becquerel/core/plotting.py
+++ b/becquerel/core/plotting.py
@@ -6,6 +6,7 @@ import numpy as np
 from uncertainties import unumpy
 from .utils import isstring
 
+
 class PlottingError(Exception):
     """General exception for plotting.py"""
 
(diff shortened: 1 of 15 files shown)


becquerel/core/plotting.py
```diff
@@ -532,3 +466,78 @@ def linthreshy(self):
         min_ind = np.argmin(np.abs(self._ydata[self._ydata != 0]))
         delta_y = np.abs(self._ydata - self._ydata[min_ind])
         return np.min(delta_y[delta_y > 0])
+
+
+def get_channel_edges(channels):
```
jccurtis: @cosama why did you remove the `staticmethods` in favor of funcs? Is there a style guideline regarding this modification?

cosama: Yeah it is in the google style guide to rather use module methods. Not sure why though. Do you have any ideas?

jccurtis: Nope. Let’s follow the google style guide on this. Thanks!

jccurtis:
We are waiting to fixing linting until after v0.2.0

jccurtis:
This branch will need to be remade (maybe a new PR?) Moving to `v1.0.0`

markbandstra:
Made obsolete by #221 and #220

Issue #112, opened 2018-05-23
Spectrum multiplication, addition and subtraction problems


cosama:
* The spectrum has no `__rmul__` method, thus `2*spectrum` produces an error, while `spectrum*2` works. This should be easy to add, it would be pretty much just a copy of `__mul__.` This doesn't affect division, I think `2/spectrum`, doesn't make sense.

* If both spectra in addition/subtraction don't have cps defined an error is thrown. I think in that case it should just ignore cps. If one has cps defined and the other does not then, maybe it would make sense to throw an error, or maybe better not define the cps in the new spectrum (which then will cause an error once this method will be accessed in the new spectrum).

cosama:
I just add this here, even though it is only slightly related. If an array with negative values is passed to the spectrum class it causes a warning in  `counts, uncs, lambda x: np.maximum(np.sqrt(x), 1))`. Not sure what to do about that.

jccurtis:
- Good catch, I've moved the negative input discussion to #113.
- The spectra sub operation doesn't make sense unless you are operating on time dependent values. Can you think of a use case in which you would subtract counts?
- The spectra add operation should be able to handle counts when no cps / livetime are defined. Is this not the case? We should add a test ...

cosama:
To be honest the reason why this all popped up is because I needed a spectrum with negative counts to test errors related to y-limits when creating log plots (which can not include negative values). The only two ways to get one is to either pass a np array with negative values to the constructor or to subtract a spectrum from the other. It is something that probably is hardly ever encountered in reality.
* The only case I could think of the subtraction being necessary without having lifetimes, would be if you know that two measurements are the same lifetime, but you don't know what it is. However, then it would be okay to just use a random value, but maybe better to allow for that.
* The add operator calls the spec.cps attribute, which throws an error if spec._cps or spec.lifetime is not defined. We could simply catch the error and not assign anything to spec._cps in the new spectrum if that happens.

cosama:
I started to work on this (in particular the right multiplication) and found the following interesting behavior after adding a `__rmul__` method:

`np.array([2,3]) * spectrum` returns a list of two spectra, the first multiplied by two, the second multiplied by 3. The reason for this is that it does use the `__mul__` method of numpy, which does internally call the multiplication method for each element of the numpy array with the spectrum class, which is the  spectrum __rmul__ method in this case).

[...]

PR #129, opened 2018-07-17
Fix right multiplication of float with Spectrum not being implemented

cosama:
Fixes part of #112.

It extends part of the spectrum class so that right multiplication of a spectrum with a float can be performed. It also makes sure that any right operation with a numpy array raises an error, some of them might have worked and created confusing results.


becquerel/core/spectrum.py
```diff
@@ -203,6 +203,11 @@ def __init__(self, counts=None, cps=None, uncs=None, bin_edges_kev=None,
         else:
             self.infilename = None
 
+        # These two lines make sure operators between a Spectrum
+        # and a numpy arrays are forbidden and cause a TypeError
+        self.__array_ufunc__ = None
+        self.__array_priority__ = 1
```
jccurtis: So this only allows scalar operations?

becquerel/core/spectrum.py
```diff
@@ -529,6 +534,9 @@ def __mul__(self, other):
 
         return self._mul_div(other, div=False)
 
+    # This line will allow the right multiplication to work
+    __rmul__ = __mul__
```
jccurtis: Shouldn't this go at the top of the class def?

tests/core_test.py
```diff
@@ -536,10 +538,16 @@ def test_basic_mul_div(spectype, factor):
...
-    assert mult.livetime is None
+    mult_left = spec * factor
+    assert np.allclose(mult_left.counts_vals, factor * spec.counts_vals)
+    assert np.allclose(mult_left.counts_uncs, factor * spec.counts_uncs)
+    assert mult_left.livetime is None
+
+    mult_right = factor * spec
+    assert np.allclose(mult_right.counts_vals, factor * spec.counts_vals)
+    assert np.allclose(mult_right.counts_uncs, factor * spec.counts_uncs)
+    assert mult_right.livetime is None
```
jccurtis: Should this be changed to `np.isnan` like below? Otherwise I find `pd.isnull` catches more `null` cases.

```python
In [11]: pd.isnull(None)
Out[11]: True

In [12]: np.isnan(None)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-12-f53008f0bd14> in <module>()
----> 1 np.isnan(None)

TypeError: ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

In [13]: pd.isnull(np.nan)
Out[13]: True
```

That said `np.isnan` is a bit faster for numeric arrays:

```python
In [20]: null_arr = np.array([np.nan, 1.] * 10000)

In [21]: %timeit np.isnan(null_arr)
4.07 µs ± 19 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [22]: %timeit pd.isnull(null_arr)
14.6 µs ± 36.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

becquerel/core/spectrum.py
cosama: I forgot that in python order doesn't matter, I can move it to the top it probably is better to have it there

becquerel/core/spectrum.py
cosama: These two attributes only govern the behavior between a custom class and numpy arrays. There is a bit more information [here](https://docs.scipy.org/doc/numpy/reference/arrays.classes.html#numpy.class.__array_ufunc__). 

lf you have something like `anyobj * myobj`, then python checks for the `anyobj.__mul__` method and only if that one returns `NotImplemented` it looks for `myclass.__rmul__`. However, `numpy.arrays()` have a generic `__mul__` method that literally works on any python object, so we have to tell numpy that it should not call that method but raise an error instead. The second line is the method used in older version of numpy, ufuncs are a more modern way.

Limiting multiplication to scalars as far as we can is implemented in the __mul__ method.

tests/core_test.py
cosama: I'm a bit confused about that? You want to check for `pd.isnull(mult_right.livetime)` or `pd.isnull(mult_right.counts_vals - factor * spec.counts_vals)`? I think at this point the `livetime` has to be `None` by construction.

becquerel/core/spectrum.py
cosama: Not sure why, it works locally, but travis doesn't like the `__rmul__` statement on top, so I moved it back down, I don't think it is worth the time figuring out why it wouldn't work on travis either.

tests/core_test.py
jccurtis: Oh right, okay this is good to go.

tests/core_test.py
```diff
@@ -550,9 +558,14 @@ def test_basic_mul_div(spectype, factor):
 def test_cps_mul_div(uncal_spec_cps, factor):
     """Multiplication/division of a CPS spectrum."""
 
-    mult = uncal_spec_cps * factor
-    assert np.allclose(mult.cps_vals, factor * uncal_spec_cps.cps_vals)
-    assert np.isnan(mult.livetime)
+    mult_left = uncal_spec_cps * factor
+    assert np.allclose(mult_left.cps_vals, factor * uncal_spec_cps.cps_vals)
+    assert np.isnan(mult_left.livetime)
```
jccurtis: Okay @markbandstra reminded me that this line differs than the content above. I will make a change to change all of these be `is None`

jccurtis: I moved this conversation to #140. I think this issue is ready for merging

PR #132, opened 2018-07-19
Add energy calibration example notebook

### LINKED ISSUE #56: Energycal example notebooks
But maybe this should be combined with peak fitting once we have that?

markbandstra:
My first quick take on this. Fixes #56


cosama:
Great work. Very close to what I would have done too.

You should call the `cal.coeff` method instead of the `cal._coeff` method. Furthermore, we could add something like `max1 = 350 + np.argmax(spec.counts_vals[350:450])` as a simple way of automating the calibration without fitting and mention something along these lines.

Are we going to name examples `_examples.ipynb` or just a name. I think the example is redundant and we should remove it at a later point from the other files. So your name would be fine.

markbandstra:
Good point, somehow I missed the property `coeffs`.

Not sure what you mean by:
> Furthermore, we could add something like `max1 = 350 + np.argmax(spec.counts_vals[350:450])` as a simple way of automating the calibration without fitting and mention something along these lines.

Agreed on naming, we can change that in the rest later.

cosama:
What I meant was, instead of assuming the peak is located at channel 400, we could use a maximum search (`numpy.argmax`) in that vicinity to get a slightly better channel value, maybe the same for the line around channel 230. It's just something that crossed my mind, not really crucial if you don't think it adds any value.

Issue #107, opened 2018-04-10
Cleanup quickstart-notebook branch and cherry-pick commits into new branches 


jccurtis:
The `quickstart-notebook` branch is now a mess of various features which should be cherry-picked onto a different branch (or branches) to clean up/isolate the PR's to merge the features.

I also think this is a VERY good time to move towards a [git-flow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) like scheme with use of the `develop` branch. The master branch would then only have PR's for a new release version which we should also start doing.

I will complete this after our meeting on Friday which will require use of this  `quickstart-notebook` branch.

Proposed cherry-picking:

overview-notebook (#137)
- [x] 5a7c787ce346f9c80569a7324ffdf6eeb0047371
- [x] 30a4cd77268740f21259d182abac7169c7ea23e8
- [x] b73388a36b4a42123b22d44d075143aafb6c8739
- [x] b0b3fa0e87f6a731df99f9b89475278816a61298
- [x] 6abf97e9a7bf1c7b86139c1c2c865d0acca2e34a
- [x] 9e3047e3a931c5457978166e4eec97e6224e15be
- [x] 138c38948a9395cd37f9020ce51bc786055fb614
- [x] 0ae396d

spectrum-repr (see #134)
- [x] 05d592d54972218435fba1ad74dfb7d95e6f8227
- [x] 814aee5ac77803632e783d8b396687c1fcac9989
- [x] 2e6c1e718d20dc00f8f580cf97ca2c5eda704b02

spectrum-plot
- NOTE: this was finished by @cosama in #114
- 3997193d1f452d6bcc06c1ffe19cd40da14dfdf5
- 02e8792bdf1e2aafff3ed293feaa1dad598f5e0f

fix-#98 (see #135)
- [x] 24566697e3dbc177cba7f3f9fbc45d5a93bbaafe

fix-#101 (see #136)
- [x] af34b9fdc122f8db1c2bf471c0b80770b6a22c17

jccurtis:
I've cherry-pick'd the commits from the now stale branch `quickstart-notebook`. I've created PR's with the relevant links above. I'll leave this open because it will be closed with #137.

Issue #94, opened 2018-03-27
Overview notebook


jccurtis:

markbandstra:
Here's a potential outline off the top of my head:

* Core
    * The `Spectrum` object
        - demonstrate how to instantiate the object with count data
        - I/O from standard file types
        - explain how uncertainties are handled
        - demonstrate various properties: e.g., `cps`, `cpskev`
        - demonstrate magic methods (add, sub, mul, div)
    * Calibrations
        - demonstrate how to instantiate a `LinearEnergyCal` and assign it to a `Spectrum`
        - calibrations are automatically created and included in a `Spectrum` when reading from a file
* Nuclear data tools
    * NNDC data
        - demonstrate how to fetch nuclear wallet card data for common isotopes, e.g., isotopes of potassium
        - demonstrate how to fetch gamma-ray lines for a common isotope, e.g., Cs-137
    * `Isotope`
        - demonstrate instantiation with, e.g., `'Carbon-12'`, `'K-40'`, and `'Tc-99m'`
        - demonstrate properties, e.g., `half_life`, `is_stable`, `decay_modes`, `j_pi`
        - demonstrate string formatting, e.g., `'{:%n(%s)-%a%m Z=%z A=%a}'.format(iso)`
* Material data tools
    * XCOM data
        - demonstrate how to fetch attenuation data for an element like Ge or Pb, and plot the results
        - same, but a compound like H2O
        - same, but a mixture like air
    * Material data
        - demonstrate element data with `materials.fetch_element_data()`
        - demonstrate compound data with `materials.fetch_compound_data()`

PR #135, opened 2018-07-20
Add SPEFile presets fields

### LINKED ISSUE #98: SPE file $PRESETS field
Reading the test spectrum (`1110C NAA cave pottery.Spe`) results in these warnings for the `$PRESETS` field:

```bash
Line 16414 unknown:  $PRESETS:
Line 16415 unknown:  Live Time
Line 16416 unknown:  86400
Line 16417 unknown:  0
```

We should added these as metadata of the `Spectrum` object as in  #81.

jccurtis:
Closes #98

The PR adds the `$PRESETS` field to the `Spe` file parser. This field is included in `Spe` files created by the ORTEC Digibase USB devices. This code was originally included in #94 and #107

TODO: 
- [x] Add warning
- [x] Add custom ParserWarning
- [ ] Confirm Travis finishes for `py27`


jccurtis:
@chunhochow brings up a good point to through a warning

markbandstra:
@jccurtis Please look into the CI build errors.

jccurtis:
@markbandstra @cosama 

Check out the fix I just pushed in 8535870 (no numbers in the hash 😨)

The configuration is from [here](https://docs.pytest.org/en/latest/reference.html#configuration-options). The specific issues is mentioned in an open issue [here](https://github.com/pytest-dev/pytest/issues/2917)

jccurtis:
Weird, I cannot produce the error on Travis locally in `py27`...

markbandstra:
I had to find out: the probability of a 7-digit hash made of only numerals is (10/16)^7 ~ 3.7%.

cosama:
The issue with travis was, the `autocal_test.py` calls `bq.Spectrum.from_file(filename2)` on line 31. As this is not wrapped in a with statement it 'eats away' the warning, and it won't be raised later on. I added a with block there too, so that it is required to raise a warning.

jccurtis:
@markbandstra yes that should be it!

jccurtis:
FYI the pytest folks just patched this behavior. Hopefully the warnings catch will be more robust between py2/3

PR #136, opened 2018-07-20
 Add Spectrum.bin_edges_keV property

### LINKED ISSUE #101: Change Spectrum.bin_edges_keV to a property
The bin_edges_keV array should be hidden: `_bin_edges_keV` and the setter should perform this work which is currently in the `__init__`:

```python
        if bin_edges_kev is None:
            self.bin_edges_kev = None
        elif len(bin_edges_kev) != len(self) + 1:
            raise SpectrumError('Bad length of bin edges vector')
        elif np.any(np.diff(bin_edges_kev) <= 0):
            raise ValueError(
                'Bin edge energies must be strictly increasing')
        else:
            self.bin_edges_kev = np.array(bin_edges_kev, dtype=float)
```

jccurtis:
Closes #101, code from #94 and #107. 

TODO:

- [x] Resolved HACK for first negative edge


jccurtis:
We should figure out the cases in other modules that cannot take negative `bin_edges` as input

jccurtis:
Related to `rebinning`

jccurtis:
@cosama:

Add to property setter
- check len of edges
- add `np.asarray(bin_edges, dtype=float)`

cosama:
That should be it. Let me know if I should change something.

PR #93, opened 2018-03-27
Autocalibration tools

markbandstra:
This commit adds some basic tools to automatically calibrate a `Spectrum` using minimal information about its gain. The notebook `autocal.ipynb` explains the process, but it amounts to:

1. Create a `PeakFilter` using a rough FWHM (in channels) at one channel, and use it to instantiate a `PeakFinder`.
2. Use `PeakFinder.find_peaks` to find the highest signal-to-noise peaks in the spectrum.
3. Instantiate an `AutoCalibrator` with a `PeakFinder`.
4. Use `AutoCalibrator.fit` with a list of energies to match to the peak candidates.

Remaining TODO items:
- [x] Split peak-finding functionality out of `AutoCalibrator` into `PeakFinder`
- [x] Use `PeakFinder` to estimate both mean and FWHM of each line.
- [x] Add tests.
- [x] Update readme for `sim_spec`
- [x] Ensure usability of the API and robustness with other real-world examples.
- [x] Organization: should `autocal` remain in `core` or move to `tools`?
- ~[ ] Integration: integrate autocalibration into a `Spectrum` method?~

PR #138, opened 2018-07-23
Add example spectrum inventory table

jccurtis:
Add information table for example spectrum files

Related to the table originally created in #93


jccurtis:
Thanks @markbandstra !

PR #144, opened 2018-08-30
Spectrum bugfixes

### LINKED ISSUE #112: Spectrum multiplication, addition and subtraction problems

### LINKED ISSUE #113: Negative inputs to Spectra
@cosama noted the a Spectra input with negative counts causes a warning when calculating uncertainty via `np.sqrt`. See: https://github.com/lbl-anp/becquerel/blob/develop/becquerel/core/spectrum.py#L126-L146

```python
        if counts is not None:
            if len(counts) == 0:
                raise SpectrumError('Empty spectrum counts')
            self._counts = handle_uncs(
                counts, uncs, lambda x: np.maximum(np.sqrt(x), 1))
            if livetime is None:
                self.livetime = None
            else:
                self.livetime = float(livetime)
            self._cps = None
        else:
            self._counts = None
            if livetime is None:
                self.livetime = np.nan  # default for CPS-based spectra
            else:
                self.livetime = livetime
                # TODO should this be allowed?
                #   all calculations with CPS return livetime=np.nan anyway...
            if len(cps) == 0:
                raise SpectrumError('Empty spectrum counts')
            self._cps = handle_uncs(cps, uncs, lambda x: np.nan)
```

This can cause confusion and doesn't gracefully handle the situation. I propose a check for negative values in the `counts` which raises an error. Thoughts?

### LINKED ISSUE #122: Allow Spectrum counts calculation from cps and livetime
As pointing out by @cosama, we currently cannot calculate `Spectrum.counts` from `Spectrum.cps` and `Spectrum.livetime`. This is not symmetric with the corresponding `Spectrum.cps` calculation. This should be a quick patch with a test to assure functionality.

### LINKED ISSUE #140: Spectrum.livetime null value
Currently a `Spectrum` can have numerical (`float`) livetime or two different `null` values if the the `livetime` was not provided. These cases can be found in `Spectrum.__init__`:

1. `None` if `counts` are provided
2. `np.nan` if `cps` are provided

This makes testing for the `livetime` a bit tricky as noted in #129:

> Should this be changed to `np.isnan` like below? Otherwise I find `pd.isnull` catches more `null` cases.
> ```python
> In [11]: pd.isnull(None)
> Out[11]: True
> 
> In [12]: np.isnan(None)
> ---------------------------------------------------------------------------
> TypeError                                 Traceback (most recent call last)
> <ipython-input-12-f53008f0bd14> in <module>()
> ----> 1 np.isnan(None)
> 
> TypeError: ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
> 
> In [13]: pd.isnull(np.nan)
> Out[13]: True
> ```
> That said `np.isnan` is a bit faster for numeric arrays:
> ```python
> In [20]: null_arr = np.array([np.nan, 1.] * 10000)
> 
> In [21]: %timeit np.isnan(null_arr)
> 4.07 µs ± 19 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
> 
> In [22]: %timeit pd.isnull(null_arr)
> 14.6 µs ± 36.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
> ```

With this in mind, @markbandstra and I agree that if the `livetime` is not provided (or is no longer relevant - say after a subtraction operation with two different `livetime`'s) then `livetime` should be set to `None` and not the `float` `np.nan`. 

I'll change this and update the tests unless I've missed something in the `core` which depends on this nuance.

cosama:
Closes #140 
Closes #122 
Closes #113 
Closes #112 

To do:
- [x] If livetime not provided it should be always `None` (#140)
- [x] Allow for counts from cps and livetime (#122)
- [x] Behavior of addition and subtraction and raise warnings when necessary (#112)
- [x] Negative input to spectrum (#113)
- [x] Fix tests for new behavior
- [x] Add new tests, remove unnecessary tests
- [x] Fix doc strings
- [x] lints, spell check

Remarks:
 * Addition: Only allowed it if both spectrum were initiated identically with `counts` or `cps`, respectively. If this is not the case the user can enforce addition with `Spectrum(counts=specA.counts+specB.counts)` or `Spectrum(cps=specA.cps+specB.cps)` as suggested in the error message.
    * Question: Should a warning be raised if `counts` were added and `livetime` have been ignored (because one of them was `None`)?
* Subtraction: If `cps` is available in both subtract `cps` and create a new spectrum with `cps`. If one of the `cps` values has been derived from `counts` and `livetime` it raises a warning. If `cps` is not available in one of the spectra try subtraction of `counts` and set `uncs` to `np.nan`; this always raises a warning.
    * Question: if `counts`-subtraction fails it fails with a ''Unknown livetime; cannot calculate counts from CPS" `SpectrumError`, should we raise a special subtraction failed error message?

### DIFF
--- a/becquerel/core/spectrum.py
+++ b/becquerel/core/spectrum.py
@@ -9,9 +9,17 @@ from uncertainties import UFloat, unumpy
 from .. import parsers
 from .utils import handle_uncs, handle_datetime, bin_centers_from_edges
 from . import plotting
+import warnings
+
 
 class SpectrumError(Exception):
     """Exception raised by Spectrum."""
+
+    pass
+
+
+class SpectrumWarning(UserWarning):
+    """Warnings displayed by Spectrum."""
 
     pass
 
@@ -123,27 +131,23 @@ class Spectrum(object):
 
         if not (counts is None) ^ (cps is None):
             raise SpectrumError('Must specify one of counts or CPS')
+
         if counts is not None:
             if len(counts) == 0:
                 raise SpectrumError('Empty spectrum counts')
             self._counts = handle_uncs(
                 counts, uncs, lambda x: np.maximum(np.sqrt(x), 1))
-            if livetime is None:
-                self.livetime = None
-            else:
-                self.livetime = float(livetime)
             self._cps = None
         else:
-            self._counts = None
-            if livetime is None:
-                self.livetime = np.nan  # default for CPS-based spectra
-            else:
-                self.livetime = livetime
-                # TODO should this be allowed?
-                #   all calculations with CPS return livetime=np.nan anyway...
             if len(cps) == 0:
                 raise SpectrumError('Empty spectrum counts')
             self._cps = handle_uncs(cps, uncs, lambda x: np.nan)
+            self._counts = None
+
+        if livetime is None:
+            self.livetime = None
+        else:
+            self.livetime = float(livetime)
 
         if bin_edges_kev is None:
             self.bin_edges_kev = None
@@ -212,40 +216,44 @@ class Spectrum(object):
     def counts(self):
         """Counts in each channel, with uncertainty.
 
-        Some spectra, including subtraction results, have cps but no counts,
-        in which case counts is None.
+        If cps is defined, counts is calculated from cps and livetime.
+        Otherwise, it is an independent data property.
+
+        Raises:
+          SpectrumError: if cps is defined, but not livetime
 
         Returns:
-          an np.array of uncertainties.ufloats, or None
+          an np.array of uncertainties.ufloats
         """
 
-        return self._counts
+        if self._counts is not None:
+            return self._counts
+        else:
+            try:
+                return self.cps * self.livetime
+            except TypeError:
+                raise SpectrumError(
+                    'Unknown livetime; cannot calculate counts from CPS')
 
     @property
     def counts_vals(self):
         """Counts in each channel, no uncertainties.
 
         Returns:
-          an np.array of floats, or None
+          an np.array of floats
         """
 
-        if self.counts is None:
-            return None
-        else:
-            return unumpy.nominal_values(self._counts)
+        return unumpy.nominal_values(self.counts)
 
     @property
     def counts_uncs(self):
         """Uncertainties on the counts in each channel.
 
         Returns:
-          an np.array of floats, or None
+          an np.array of floats
         """
 
-        if self.counts is None:
-            return None
-        else:
-            return unumpy.std_devs(self._counts)
+        return unumpy.std_devs(self.counts)
 
     @property
     def cps(self):
@@ -418,9 +426,9 @@ class Spectrum(object):
           an int
         """
 
-        if self.counts is not None:
+        try:
             return len(self.counts)
-        else:
+        except SpectrumError:
             return len(self.cps)
 
     def __add__(self, other):
@@ -438,7 +446,8 @@ class Spectrum(object):
         Raises:
           TypeError: if other is not a Spectrum
           SpectrumError: if spectra are different lengths,
-            or if only one is calibrated
+            if only one is calibrated or if specra have not both
+            been initialized with counts or CPS, respectively.
           NotImplementedError: if spectra are calibrated differently
 
         Returns:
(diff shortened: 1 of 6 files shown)


becquerel/core/spectrum.py
```diff
@@ -212,40 +216,44 @@ def __init__(self, counts=None, cps=None, uncs=None, bin_edges_kev=None,
...
-          an np.array of uncertainties.ufloats, or None
+          an np.array of uncertainties.ufloats
         """
 
-        return self._counts
+        if self._counts is not None:
+            return self._counts
+        else:
+            try:
+                return self.cps * self.livetime
```
cosama: Maybe this should be
```
if self.livetime is not None:
    return self._cps * self.livetime
else:
    raise ...
```

becquerel/core/spectrum.py
```diff
@@ -471,19 +485,32 @@ def __sub__(self, other):
...
-        spect_obj = Spectrum(cps=cps, bin_edges_kev=self.bin_edges_kev)
-
+        try:
+            kwargs = {'cps': self.cps - other.cps}
+            if (self._cps is None) or (other._cps is None):
+                warnings.warn('Spectrum subtraction of counts, spectra ' +
+                              'have been converted to CPS', SpectrumWarning)
+        except:
+            kwargs = {'counts': self.counts_vals - other.counts_vals}
+            kwargs['uncs'] = [np.nan]*len(self)
```
cosama: Maybe that is not necessary. Are the uncertainties in this case meaningful?

becquerel/core/spectrum.py
```diff
@@ -438,16 +451,22 @@ def __add__(self, other):
...
 
         self._add_sub_error_checking(other)
+        if (self._counts is None) ^ (other._counts is None):
+            raise SpectrumError(
+                'Spectrum addition with a mix of counts and CPS is ' +
+                'ambiguous, use Spectrum(counts=specA.counts+specB.counts)' +
+                ' or Spectrum(cps=specA.cps+specB.cps) instead.')
 
-        if self.counts is not None and other.counts is not None:
+        if self._counts is not None and other._counts is not None:
```
cosama: PyLint complains: 'Access to a protected member _counts of a client class (protected-access)'. If that is an issue we need a method to check for counts vs cps based spectra.

becquerel/core/spectrum.py
```diff
@@ -123,27 +124,29 @@ def __init__(self, counts=None, cps=None, uncs=None, bin_edges_kev=None,
 
         if not (counts is None) ^ (cps is None):
             raise SpectrumError('Must specify one of counts or CPS')
+
         if counts is not None:
             if len(counts) == 0:
                 raise SpectrumError('Empty spectrum counts')
+            if uncs is None and np.any(np.array(counts) < 0):
```
cosama: Is there a more elegant way than converting counts to a `numpy` array?

becquerel/core/spectrum.py
cosama: PyLint complains: 'Access to a protected member _counts of a client class (protected-access)'. If that is an issue we need a method to check for counts vs cps based spectra.

cosama: Not really outdated.

becquerel/core/spectrum.py
jccurtis: Counts needs to be converted to an array later anyways so I don’t think this is an issue. We should probably cast the counts array with np.asarray

jccurtis:
Yes to both questions in the description

becquerel/core/spectrum.py
cosama: I think it is fine the way it is.

cosama:
As requested by @jccurtis I added the issues raised in the two questions. The only thing that is still outstanding to me at this point is the access to a protected member by `other._counts` (see outdated review above).

becquerel/core/spectrum.py
markbandstra: I think it's okay since `other` is also a `Spectrum`.

Issue #50, opened 2017-04-24
Fitting module


bplimley:
For peak fitting, energy cal fitting.
To use `lmfit`.
@jccurtis to take code from GRPP as a start.

(see also #22)

@jccurtis TODO:
- [ ] Update `erf` naming
- [ ] Add warning when fit errors unknown

cosama:
I was looking through some paper about maximum likelihood and Poisson statistics in peak fitting. I just dump this here for now, might be useful at some later point:

- Chi-square vs maximum likelihood compared: https://www.sciencedirect.com/science/article/pii/S0168900200007567
- Maximum likelihood and small signals: https://www.sciencedirect.com/science/article/pii/S0168900299002697
- Very general article about the topic: https://www.nature.com/articles/nmeth0510-338
- Different methods to perform a likelihood fit (also iterative): https://link.springer.com/article/10.1007%2Fs10909-014-1098-4

jccurtis:
I've been building the `fitting` module in the `feature-fitting-#50` branch. I have 5 peak models so far based on the ones I have used most commonly. I still have a long list of inline TODO's (mostly short) which I'm churning through as I write an example notebook and tests. Some of these items are topics we should discuss next week:

- [ ] Handle channel / energy fitting (what are the bin edges...)
- [ ] How to handle bin width normalization. This is particularly important for calculating the are of a gaussian. In the past I have tracked the "mode" or "units" for the y data and always tried to fit in bin-width normalized units. This was troublesome so I would rather have this be more flexible (and explicit because this is a very common error I have observed people make while learning more about peak fitting).
- [ ] Plotting options
- [ ] Peak presets: I have multi gaussian models but we could have subclasses with constrained parameters for commonly fit/complex peaks in scintillators (Co57, Co60, etc)
- [ ] Parameter guessing, the parameter initialization currently occurs based on the data vectors but some improved guess work regarding the energy resolution could be very useful. I would like to use some of the low level tools from @markbandstra's auto-calibration toolkit to do this. A general function to guess the energy resolution for a vector with a peak-like feature in it could be very useful. I could do this with `scipy`'s `cwt` transform but I would rather leverage the auto-calibration functions if possible.

Eventually we should discuss a full calibration routine which uses #93 and #50 but we should get all the underlying functionality working first.

cosama:
[...]

PR #146, opened 2018-09-11
Feature fitting #50

jccurtis:
For discussion. Still requires API improvements and additional functionality based on feedback. 

- [x] Add example notebook
- [ ] Update plotting
- [ ] add docs
- [ ] add ability to override defaults
- [ ] add ability to initialize and fit with Fitter.__init__
- [x] change to use ABC
- [ ] include x_edges?
- [ ] handle y normalization (i.e. cps vs cps/keV), needs x_edges
- [ ] use set_param_hint to set global model defaults
- [x] Update overview notebook


cosama:
Just a mental note to myself. We should be using spectrum.plot in the notebook, also there is a semi column sitting there.

cosama:
I fixed the travis build issue with python2 for now.

@jccurtis Thanks for bringing this so far. I would say overall it looks pretty good.

In addition to what @jccurtis mentioned above, I would also like to add a few things that I think should go in before we can merge it (probably more to come):
- [ ] Add FitterGauss and FitterGaussLine
- [ ] Fix notbook issues mentioned above
- [ ] Check weights (I think we don't need to have 1/weight)
- [ ] Add `fit_report` or `report` function to `Fitter` class
- [ ] Add fitter method to `Spectrum` class
- [ ] More examples in notebook

jccurtis:
Note to self: remove need for units in fitter

markbandstra:
Notes from today:
- [ ] Testing needs to be expanded to test all models
- [ ] Gaussian model by area in addition to amplitude (or eliminate by amplitude?) -- this reduces covariance between area and width, and also is in general easier to interpret area
- [ ] Decide what to do about unequal bin width -- raise an exception?
- [ ] Be clear about primary use cases: fitting count-channel histograms to find peaks for energy calibration(?)
- [X] Return fit uncertainties
- [ ] Add reference to Bevington in documentation -- Lorentzian peak on quadratic background (@jvavrek )

becquerel/core/fitting.py
```diff
@@ -0,0 +1,537 @@
...
+    return np.ones_like(x) * c
+
+
+def line(x, m, b):
+    return m * x + b
+
+
+def gauss(x, amp, mu, sigma):
+    amp = np.abs(amp)
+    mu = np.abs(mu)
```
jvavrek: Is there any reason to restrict `mu` to non-negative values? It's probably usually true for spectral peaks, but this feels more restrictive than robust.

becquerel/core/fitting.py
```diff
@@ -0,0 +1,537 @@
...
+
+
+def erf(x, amp, mu, sigma):
+    return amp * 0.5 * (1. - scipy.special.erf((x - mu) / (1.414214 * sigma)))
+
+
+def exp(x, amp, lam):
+    return amp * np.exp(x / lam)
+
+
```
jvavrek: For some reason, I can't find it in this diff, but there are some undefined names in `def expgaussian()` on what I've pulled to my machine.

becquerel/core/fitting.py
jccurtis: Good point. I suppose not necessarily. This should be a higher level configuration specifically for peak fitting in the constraints.

jccurtis:
@jvavrek somehow there are a bunch of extra commits from other (merged?) branches on this branch now. How did you merge this branch? We may need to scrap this branch and cherry pick since this is a bit messy...

jccurtis:
@jvavrek specifically caff381

jccurtis:
@Jayson Vavrek I might force push to fix this using my local branch and then cherry pick your commit (0a906ae). Let me know if you are making any other changes. @cosama do you have any unpushed code?

jvavrek:
Oof, I see all those commits now. It's likely something went awry when I checked out the branch. Let me know if I can help fix it. I only wanted to add the docstring, so feel free to revert what I've done.

jccurtis:
I think what happened is that you merged your local develop branch with develop without the commit I made today for the random seed (see [network](https://github.com/lbl-anp/becquerel/network)). No worries though. I will fix it. 😄

jvavrek:
@jccurtis Looks like I inadvertently did `git checkout -b` instead of just `git checkout`, which would create a new branch with the same name but base it off of `develop`

jccurtis:
Ah okay, I also merged develop and pushed during the meeting (I forgot about that). Looking a bit more at the diff, lets just keep moving and I won't take any action. The current state isn't destructive so lets just focus on getting these items finished. Thanks for the reference! I should really check that book out ...

jccurtis:
Ack, well okay the diff online is completely broken as a result and doesn't recognize the merge commit. Attempting a fix. Hold tight!

jccurtis:
@jvavrek fixed! Delete your local branch and checkout again

cosama:
With [`scipy.optimize.golden`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.golden.html) and [`scipy.optimize.bisect`](https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.optimize.bisect.html) it should be possible to find the FWHM of any 'peak shaped' function. Is this something we would like to implement to have a general way of calculating that?

Maybe there is a better way?

cosama:
I fixed the exponentially modified Gaussian, it should work now, parameter estimation is not perfect, but probably good enough for now.

Instead of adding models with `area` all over, couldn't we just add a parameter `area` to the models that converts from `amp` to `area`. `lmfit` seems to be laid out to allow for this specifically and we already do it for `fwhm`.

cosama:
Just a quick code snipped to get the fwhm of any function, here as an example for the `expgauss` function:

```
    amp, mu, sigma, gamma = 10000, 100, 5, 0.25
    minimum = scipy.optimize.golden(expgauss, args=(-amp, mu, sigma, gamma))
    ma = expgauss(minimum, amp, mu, sigma, gamma)
    left = scipy.optimize.brentq(lambda x : expgauss(x, amp, mu, sigma, gamma) - ma*0.5,  0, minimum)
    right = scipy.optimize.brentq(lambda x : expgauss(x, amp, mu, sigma, gamma) - ma*0.5, minimum, 200)
    print(right - left)
```

The code snipped also shows that the fwhm is not equal to `sigma*sqrt(8*log(2))` in the case of an exponentially modified gaussian, as assumed here: https://github.com/lmfit/lmfit-py/blob/master/lmfit/models.py#L883

markbandstra:
It could be useful to provide @cosama 's `calc_fwhm` function as a standalone convenience function for the user. It might also be useful to introduce a class inheritance structure that distinguishes peak models from background models and provides the function as a method for the former kind.

cosama:
All tests passed now. Issue was `np.random.seed` call at top of file, which sets the seed on initialization.  Results of successive calls to `np.random` will depend on how many time it has been called in other tests before.

cosama:
Does `line` model need a constant offset? It would be easy to have that through `constant`.

jccurtis:
- [ ] add option to include weights or not in cases which you don't have per bin uncertainty

becquerel/core/fitting.py
jccurtis: is this still the case?

jvavrek: No, looks like all names are defined now.

becquerel/core/fitting.py
```diff
@@ -0,0 +1,681 @@
...
+def line(x, m, b):
+    return m * x + b
+
+
+def gauss(x, amp, mu, sigma):
+    amp = np.abs(amp)
+    mu = np.abs(mu)
+    sigma = np.abs(sigma)
+    return amp / np.sqrt(2. * sigma ** 2. * np.pi) * \
+        np.exp(-(x - mu) ** 2. / (2. * sigma ** 2.))
```
jccurtis: @markbandstra check this out!

jccurtis:
Hi everyone, to follow up on our earlier conversation, the current Gaussian formulation is a density function:

gauss=\frac{A}{\sigma\sqrt{2\pi}}\exp\left(-\frac{(x-\mu)^2}{2\sigma^2}\right))

The point I was making over the call regarding units was:

* x in keV
* if y in cnts
  * A in cnts/keV
* if y in cnts/keV (density of cnt data)
  * A in cnts

The second case corresponds to the the formulation of the Gaussian function above and `A` is the integrated area.

My confusion comes from not understanding how to use the first case to find the area. I suppose you could take the sum of `y(x_input)` to get the peak area. I don't really understand how the Poisson fitting of the Gaussian would work given this formulation. Shouldn't one use the non-density formulation of a Gaussian:

gauss=A\exp\left(-\frac{(x-\mu)^2}{2\sigma^2}\right))

Any additional thoughts would be extremely helpful to resolve this and get this issue / PR completed. Thanks!

cosama:
I think where this should be going (and I think @markbandstra was referring to this is full Poisson likelihood fit). Here a example how this would need to be done in the case of linear and sqrt binned histograms:

```
from scipy.optimize import minimize
import numpy as np
import matplotlib.pyplot as plt

def gauss(x, amp, mu, sigma):
    amp = np.abs(amp)
    mu = np.abs(mu)
    sigma = np.abs(sigma)
    return amp / np.sqrt(2. * sigma ** 2. * np.pi) * \
        np.exp(-(x - mu) ** 2. / (2. * sigma ** 2.))


def fit(model, counts, edges):
    def PMLE(params, model, counts, edges):
        y = model((edges[1:] + edges[:-1])*0.5, *params)*(edges[1:] - edges[:-1])
        return 2 * np.sum(y - counts * np.log(y))  # + some constant (function of counts)

    def guess(counts, edges):
        centers = (edges[1:] + edges[:-1])*0.5
        sum_cnts = np.sum(counts)
        mean_cnts = np.sum(counts * centers) / sum_cnts
        sdtdev_cnts = np.sqrt(np.sum(counts * (centers - mean_cnts)**2) / sum_cnts)
        return [sum_cnts, mean_cnts, sdtdev_cnts]  # this might be to good in this case

    params = guess(counts, edges)
    print("Inital guess", params)
    return minimize(PMLE, params, args=(model, counts, edges))

amp, mu, sigma = 10000, 100, 20
print("Initial values", amp, mu, sigma)

# Linear bins
fedges = np.arange(0, 256)
centers = (fedges[1:] + fedges[:-1]) * 0.5
widths = (fedges[1:] - fedges[:-1])
fdata = np.random.poisson(gauss(centers, amp, mu, sigma)*widths)
print("Sum events linear", np.sum(fdata))

res = fit(gauss, fdata, fedges)

print("Result linear", res.x)
plt.figure()
plt.step(centers, fdata, where='mid')
plt.plot(centers, gauss(centers, *res.x)*widths)

# Sqrt bins
fedges, _ = sqrt_n_bins(nmin=0, nmax=256, nbins=20)
centers = (fedges[1:] + fedges[:-1]) * 0.5
widths = (fedges[1:] - fedges[:-1])
fdata = np.random.poisson(gauss(centers, amp, mu, sigma)*widths)
print("Sum events sqrt", np.sum(fdata))

res = fit(gauss, fdata, fedges)

print("Result sqrt", res.x)
plt.figure()
plt.step(centers, fdata, where='mid')
plt.plot(centers, gauss(centers, *res.x)*widths)
```
In this case `counts` needs to stay counts, so we need to normalize the model by multiplying it with the bin width. I don't think this is something that we currently support using `lmfit`. It also resolves the units questions as far as I can tell, the user has to assure that it is in `counts`. I think for `cnts/sec` we could use an analog process but use a Pearson or Newman chi2 in PMLE.

markbandstra:
Here is my conception of what is going on, I think it is close to what both of you are saying.

We're talking about a count density function `f` that has units of counts per keV:

<a href="https://www.codecogs.com/eqnedit.php?latex=\large&space;f(x_i)&space;=&space;\frac{A}{\sqrt{2&space;\pi}&space;\sigma}&space;\exp\left(-\frac{(x_i&space;-&space;\mu)^2}{2&space;\sigma^2}\right)" target="_blank"></a>

## Case 1: `y` are counts
For the first case, we fit `f` to data `y` which is a series of counts `n`. Since `f` is a count density, we need to consider the (potentially non-uniform) bin width:

<a href="https://www.codecogs.com/eqnedit.php?latex=\large&space;n_i&space;=&space;y_i&space;\approx&space;f(x_i)&space;\Delta&space;x_i" target="_blank"></a>

This formulation has the advantage that `A` is approximately the sum of the counts:

<a href="https://www.codecogs.com/eqnedit.php?latex=\large&space;\sum_i&space;y_i&space;\approx&space;\sum_i&space;f(x_i)&space;\Delta&space;x_i&space;\approx&space;\int&space;f(x)&space;dx&space;\approx&space;A" target="_blank"></a>

## Case 2: `y` are counts/keV
In the second case we were fitting instead to a quantity `y` that is in units of counts/keV, so we would fit `f` to `y` without using the bin widths (this is what the current code does):

<a href="https://www.codecogs.com/eqnedit.php?latex=\large&space;\frac{n_i}{\Delta&space;x_i}&space;=&space;y_i&space;\approx&space;f(x_i)" target="_blank"></a>

But now if we wanted to interpret `A`, it is the sum of the `y`s but only when we include the bin widths in the sum:

<a href="https://www.codecogs.com/eqnedit.php?latex=\large&space;\sum_i&space;y_i&space;\Delta&space;x_i&space;\approx&space;\sum_i&space;f(x_i)&space;\Delta&space;x_i&space;\approx&space;\int&space;f(x)&space;dx&space;\approx&space;A" target="_blank"></a>

## Possible resolution
So in all of these cases `A` are the integrated counts in the gaussian and `f` has units of counts/keV (since sigma must have units of keV). But `y` has different units. For case 1, we need the bin widths but for case 2 we do not. @jccurtis is right that the fitter needs to be somehow aware of the units of `y`. I propose that we always fit the following relationship:

<a href="https://www.codecogs.com/eqnedit.php?latex=\large&space;y_i&space;\approx&space;f(x_i)&space;\Delta&space;x_i" target="_blank"></a>

But if `y` is counts/keV (Case 2), then the user should simply provide bin widths that are unity. But `A` is not in units of counts/keV but in counts due to the argument above.

This situation would be compatible with what we already have in the code, and it would be compatible going forward when we implement Poisson maximum likelihood.

I think this proposal could easily be accommodated using what we currently have by adding an optional `dx` kwarg and member to `Fitter`, e.g.,

```python
def set_data(self, y, x=None, y_unc=None, dx=None):
    # Set dx data
    if dx is None:
        self._dx = np.ones_like(self.x)
    else:
        self._dx = np.asarray(dx)
```

Then any place where the data are fit to the model, for now use `self.y / self.dx`, although in the future we would want to fit `self.y` with `self.model.eval(x, params) * self.dx`. Also `y_unc` would be affected by this.

These changes would naturally fit the current notation and obviate the need to rework everything to use histogram notation as @cosama suggested above.

jccurtis:
Thanks for your comments @cosama and @markbandstra !

Honestly this gets at the core of what I was pointing out which is:

```python
self.model.eval(x, params)  # is a density (i.e. cnts per bin width unit)
```

Thus fitting that directly to `y` in counts doesn't make sense because the units are incorrect.

The answer to my confusion is really:

```python
# for y in counts as needed for Poisson fitting:
self.y ≈ self.model.eval(self.x, params) * self.dx
```

which is a bit 🤦‍♂ on my part ...

@markbandstra's implementation of `set_data` is essentially what I had used previously (I just included units to enforce a coded sanity check for myself since the simulation validation was such a headache. To avoid confusion I would set `self.dx=None` and then don't use it when fitting but probably raise a warning if using a Gaussian. I can add that soon and write up the associated docs.

cosama:
I think we all agree on this. I had a closer look at `lmfit` and it should be straight forward to implement additional fitting methods (such as Poisson) with setting up minimizer manually as explained here: https://github.com/lmfit/lmfit-py#parameters-and-fitting. For now we can just pass in `self.y/self.dx`.

becquerel/core/fitting.py
```diff
@@ -0,0 +1,681 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super(ErfModel, self).__init__(erf, *args, **kwargs)
+
+    def guess(self, y, x=None):
+        return [
+            ('{}amp'.format(self.prefix), 'value', y[0] - y[-1]),
+            ('{}amp'.format(self.prefix), 'min', 0.),
+            ('{}mu'.format(self.prefix), 'expr', 'gauss_mu'),
+            ('{}sigma'.format(self.prefix), 'expr', 'gauss_sigma'),
```
cosama: Is there a way to generalize this? I.e. what if there are two Gaussians, or if we like to fit ExpGauss + Erf?

jccurtis: the names are generalized. are you talking about the guess values?

cosama: Sorry, wasn't too precise here. I talk about `erf_mu` being intialized as `gauss_mu` and the same for `sigma`. If you have a ExpGauss + Erf model, there is no parameter `gauss_mu`, so a call like `Fitter(['expgauss', 'erf'])` or simply `Fitter(['erf'])` fails.

I don't think we can make this more general, I think however we should be careful when we define parameters (using `make_params`, or `set_params`), so that the user can fix a conflict before something fails.

becquerel/core/fitting.py
```diff
@@ -455,9 +455,37 @@ def fit(self, backend='lmfit'):
...
+            diff = (scipy.special.xlogy(data, data) -
+                    scipy.special.xlogy(data, model) + model - data)
+            if diff.dtype == np.complex:
+                # data/model are complex
+                diff = diff.ravel().view(np.float)
+            if weights is not None:
+                raise ValueError("Weights are not supported in Poisson " +
+                                 "Likelihood minimization.")
+            return np.asarray(diff).ravel()  # for compatibility with pandas.Series
+        self.model._residual = _likelihood_residual.__get__(self.model, Model)
```
jccurtis: @cosama can you add a comment here about this overwriting the object's method

becquerel/core/fitting.py
```diff
@@ -455,9 +455,37 @@ def fit(self, backend='lmfit'):
                                          self.params,
                                          x=self.x_roi,
                                          weights=weights)
+        elif backend.lower().strip() == 'lmfit-pml':
```
jccurtis: is here a better name?

jccurtis:
@cosama and I just discussed:

@cosama:
- [ ] add some more tests
- [ ] test sum of events in peak

@jccurtis:
- [ ] make model names more specific ( i.e. gauss + gauss + erf )
    * @jccurtis check if `_` (underscores) are allowed in model name and try add that functionality
    * maybe just remove underscores from model names
- [ ] remove complications for now
    * put these in a higher level module
- [ ] add `gausserf` model
    - [ ] erf doesn't look for gauss params

higher level structure
* fit NaI peaks 
* Compton edge
* ... ?

cosama:
I added some more tests and fixed a few things that bothered me. New thing to think about:

* [ ] Calculating covariance (`lmfit` with `numdifftools`) changes optimal values. Right now stderr are not calculated for Poisson Likelihood. If that is fine we can leave this.

jccurtis:
- [ ] param uncertainty estimation (@markbandstra ) - @cosama make MWE
- [ ] add tests for `dx` @cosama  @jccurtis
- [ ] update plotting for `dx` and fix axis
- [ ] name the `lmfit-pml` backend ? @cosama @markbandstra @jvavrek 
- [ ] Add fitting into spectrum (delayed for the moment)
- [ ] Add verbose mode while fitting

jccurtis:
From slack conversation: https://iopscience.iop.org/article/10.1088/0004-637X/799/2/226/pdf

jccurtis:
Also from slack: http://adsabs.harvard.edu/full/1979ApJ...228..939C7

jvavrek:
I started trying this out for our APSL paper as a Friday afternoon project. A few things I noticed, many of which are probably just due to my unfamiliarity with this module:

- [ ] the example notebook `fitting.ipynb` is out of date. It tries to fit a `FitterGaussErfExp` but this class does not exist in `fitting.py`
- [ ] I was confused about the difference between a `Model` and a `Fitter`. The latter's docstring says 'Base class for more specialized fit objects' but its lineshapes (e.g. `GaussGaussLine`) are not substantially more complicated than those that use `Model` (e.g. `GaussErfModel`)
- [ ] similarly it looks like the two interfaces are quite different. `Fitter` has `set_roi()` but `Model` doesn't. Does the user need to manually slice the data passed to `Model`?
- [ ] it might be worth going all the way and generalizing the double Gauss models to an arbitrary number of Gaussians. I've had good results with eight Gaussians and an exponential background, e.g., so let me know if we should collaborate here
- [ ] no `LineGauss` model?
- [ ] some of the names threw me off for a while. We have `GaussErfModel` and `ExpGaussModel`; we should pick a consistent order, e.g. `[Bkg0][Bkg1]...[Signal0][Signal1]Model` or the reverse. Later I noticed that `ExpGauss` is not actually a Gaussian plus exponential background, but `GaussErf` *is* a Gaussian plus error function background
- [ ] as a result it looks like some more documentation could go a long way. For instance, are `Fitter`s just made up of `Model`s? Is that how the user should create their own (arbitrarily complex) lineshape fit?

cosama:
@jvavrek Thanks for the feedback this is certainly useful.

Here just a few replies. Models define what you want to fit, Fitter is what takes a model and turns it into a Fit, I forgot we have different fitter classes, I think we can just remove those. The easiest way to fit at this point is:

```
fitter = bq.fitting.Fitter(['gauss', 'line', 'slope'], y=ydata, x=xdata, y_unc=y_unc_data)
fitter.fit()
```

This will create internally a `GaussModel() + LineModel() + SlopeModel()` and do all the heavy juggling for you. The Models can be used to do more complex fitting, like a double gaussian that can not be done with a simple list.

`ExpGaussModel` is the Exponentially Modified Gaussian, while `GaussErfModel` is a special model that is often used with a gaussian peak and an error function with the sigma and mean fixed to each other. The order is chosen to make that clear.

GaussNModel would be could :smile: . Documentation is urgently  required, I agree.

Hope that helps some.

jvavrek:
Thanks @cosama, that helps a lot. One other QOL thing I just noticed—residuals in `Fitter.custom_plot()` might be more useful if they were relative rather than absolute.

cosama:
Here another code example that I just found, fitting two peaks, that have parameters dependent on each other:
```
model = (bq.fitting.GaussModel(prefix='gauss0_') +
         bq.fitting.GaussModel(prefix='gauss1_') +
         bq.fitting.ExpModel(prefix='exp_'))

fitter = bq.Fitter(model, y=ydata, x=xdata, y_unc=np.sqrt(ydata))
fitter.params['gauss1_mu'].set(expr='gauss0_mu*583.187/609.32')
fitter.params['gauss1_sigma'].set(expr='gauss0_sigma')
fitter.params['gauss0_sigma'].set(value=3.0)

**Comment by jccurtis** [comment]
@cosama brings up a good point. We should have common prebuilt fits in the module. If nothing else, in the notebook. I have triple peak fitting (so-so quality with NaI) for Co-57 I can upload at some point.

**Comment by jccurtis** [comment]
@jvavrek and sorry about the outdated notebook. As you can see this feature is hilariously overdue. Thanks for pushing ahead with this!

**Comment by cosama** [comment]
- [ ] Think about if there is a way of plotting `Gauss` and `Erf` independently for `GaussErf` model.

File: becquerel/core/fitting.py [hunk 7]
```diff
@@ -0,0 +1,806 @@
...
+        mu = x[0] + xspan * center_ratio
+        msk = ((x >= (mu - xspan * width_ratio)) &
+               (x <= mu + xspan * width_ratio))
+        amp = np.sum(y[msk]/dx[msk])
+        sigma = xspan * width_ratio / 10.
+        return [
+            ('{}ampgauss'.format(self.prefix), 'value', amp*amp_ratio),
+            ('{}ampgauss'.format(self.prefix), 'min', 0.0),
+            ('{}amperf'.format(self.prefix), 'value', amp*(1.0 - amp_ratio)),
+            ('{}amperf'.format(self.prefix), 'min', 0.0),
```
cosama, line 203: The `erf` function should allow for a negative amplitude.

jccurtis: You can change the value but this default comes from most cases where you want the `amp>=0`.

jvavrek:
I've found some sort of conflict between using `tex` in `matplotlib` and using the `fit.custom_plot()` method. I believe adding this line to a jupyter notebook with matplotlib in notebook/inline mode is enough to cause the error:

```
plt.rc('text', usetex=True)
```

The full error message is pretty unwieldy and doesn't quote any user code, sadly, but the root cause seems to be:

```
RuntimeError: latex was not able to process the following string:
b'gauss_'
```

in the plot legend. If I comment out or change that line, it fails next with

```
RuntimeError: latex was not able to process the following string:
b'    # fitting method   = leastsq'
```

It looks like matplotlib is having trouble with bytestrings in particular—can we change these under the hood, or plot with different encodings, for more flexibility?

jvavrek:
One minor thing—in the fit report, we should probably quote percent errors as absolute (unsigned) values. Currently I have a negative sloped `Line` background that shows up with a negative relative error on it. Add an `np.abs` here

```
s += '    {:24}: {: .6e} +/- {:.5e} ({:6.1%})\n'.format(
                    param_name, v, e, np.abs(e / v))
```

jvavrek:
@cosama @jccurtis I've made some updates:

- an option for relative residuals in the residual plot below the fit plot
- a horizontal dashed line at 0 in the residuals plot
- updates to the fit model in the fitting notebook so that it runs out of the box

Here's the example with the relative residuals:
image

I also wanted to check if the gauss amplitude guess is being set correctly. It looks like it's being set based on the sum of all counts in the ROI in `GaussModel.guess()`. Why not just the max of the ROI? That should reduce the impact of the background counts.

cosama:
Yeah the `GaussModel` and `GaussErfModel` guesses are not great. I have also seen that the amplitude is usually quite a bit to large. Note, the amplitude is not the height, it is the total number of counts in the peak. It is better to start of with a slightly larger value though. I think setting it to the max would be wrong, but we might want to divide the initial guess by the bin width maybe?

jvavrek:
This is not running on Travis, but tests pass on my machine (except for the NNDC error fixed in another MR)

cosama:
@jvavrek I just updated with `main` that should fix the `NNDC` error. Yeah travis is super slow these days, it might take a couple of hours until the job runs.

jvavrek:
> Yeah the `GaussModel` and `GaussErfModel` guesses are not great. I have also seen that the amplitude is usually quite a bit to large. Note, the amplitude is not the height, it is the total number of counts in the peak. It is better to start of with a slightly larger value though. I think setting it to the max would be wrong, but we might want to divide the initial guess by the bin width maybe?

Oh true true, I'm thinking max times sqrt(2pi) sigma (the area / height ratio is for a Gaussian). And the current version does indeed divide by the binwidth. My version
```
amp = np.max(y[msk]) * np.sqrt(2*np.pi) * sigma
```
works well and produces an initial guess much closer to the final result, but we can address that in another issue.

jvavrek:
> @jvavrek I just updated with `main` that should fix the `NNDC` error. Yeah travis is super slow these days, it might take a couple of hours until the job runs.

Cool, the NNDC error is fixed on my test suite.

becquerel/core/fitting.py
```diff
@@ -0,0 +1,816 @@
...
+
+def line(x, m, b):
+    return m * x + b
+
+
+def gauss(x, amp, mu, sigma):
+    amp = np.abs(amp)
+    mu = np.abs(mu)
+    sigma = np.abs(sigma)
+    return amp / np.sqrt(2. * sigma ** 2. * np.pi) * \
```
cosama, line 43:
```suggestion
    return amp / sigma / np.sqrt(2. * np.pi)` * \
```

jvavrek: Agreed. Is there a strong reason for `abs`ing all the inputs?

cosama: I think that is so that the optimization can't make them negative.

jvavrek: I suppose in general one could want to fit a peak with a negative mean (it's only a coordinate shift, after all) or perhaps even a trough with negative amplitude. We could probably revisit this later though. There might be a way to bound the search similar to providing initial guesses in `lmfit`, rather than bounding the fit functions themselves.

jvavrek: See also @markbandstra's comment above on parameter bounds

cosama: Yeah forcing the mean to be positive doesn't make any sense. Negative amplitudes I am not sure, often you don't want that, but there are ways to enforce that in the fitting routine

jvavrek:
@cosama @jccurtis @markbandstra are there any open comments above that are necessary to get this MR in sooner rather than later? Or can we port (some of) them to a new issue?

cosama:
I would remove the `abs`. I just check and `GaussModel` and `GaussErfModel` both set the parameter bounds to not allow negative values. I think that should be sufficient. Just need to make sure it passes all the tests.

Am I the only one who got two figure each time I call the `custom_plot` method in jupyter-notebook?

jvavrek:
> I would remove the `abs`. I just check and `GaussModel` and `GaussErfModel` both set the parameter bounds to not allow negative values. I think that should be sufficient. Just need to make sure it passes all the tests.
> 
> Am I the only one who got two figure each time I call the `custom_plot` method in jupyter-notebook?

It only plots once if you put a semicolon at the end of the line or you call `plt.show()` immediately after.

jvavrek:
> I would remove the `abs`. I just check and `GaussModel` and `GaussErfModel` both set the parameter bounds to not allow negative values. I think that should be sufficient. Just need to make sure it passes all the tests.
> 
> Am I the only one who got two figure each time I call the `custom_plot` method in jupyter-notebook?

Looks good without the `abs` to me

becquerel/core/fitting.py
```diff
@@ -0,0 +1,816 @@
...
+
+class ConstantModel(Model):
+
+    def __init__(self, *args, **kwargs):
+        super(ConstantModel, self).__init__(constant, *args, **kwargs)
+        # TODO: remove this min setting?
+        self.set_param_hint('{}c'.format(self.prefix), min=0.)
+
+    def guess(self, y, x=None, dx=None, num=2):
+        return []
```
cosama, line 91:
```suggestion
        if dx is None:
            dx = np.ones_like(x)
        c = (y[-1]/dx[-1] + y[0]/dx[0]) * 0.5
        return [
            ('{}c'.format(self.prefix), 'value', c),
        ]
```

cosama:
@jvavrek Yeah lets remove the abs then. Also can you get the change I suggested in, otherwise `gauss + const` works not very well at all.

becquerel/core/fitting.py
```diff
@@ -0,0 +1,537 @@
...
+
+
+def line(x, m, b):
+    return m * x + b
+
+
+def gauss(x, amp, mu, sigma):
+    amp = np.abs(amp)
+    mu = np.abs(mu)
+    sigma = np.abs(sigma)
```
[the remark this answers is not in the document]
jvavrek: removed

jvavrek:
@cosama looks good to me. I think we've got the important parts ready to go, so I'm ok with merging. Looks like @jccurtis is out today, but maybe @markbandstra can also weigh in?

cosama:
@jvavrek Can you try to get ride of all warnings in your notebook? The becquerel once might be a bit tricky but there might be a flag. Also I would use markdown cells instead of comments in the cells themself. Should be an easy fix. Let me know if you have questions.

jvavrek:
> @jvavrek Can you try to get ride of all warnings in your notebook? The becquerel once might be a bit tricky but there might be a flag. Also I would use markdown cells instead of comments in the cells themself. Should be an easy fix. Let me know if you have questions.

I've just deleted them from the .ipynb source if that's ok. Agreed on markdown.

jvavrek:
Actually, we probably want the residuals to be
```
(y_eval - self.y_roi) / np.abs(y_eval)
```
instead of
```
(y_eval - self.y_roi) / self.y_roi
```
right?

cosama:
I would say `y_eval - self.y_roi) / self.y_uncs_roi` or something along these lines, if y_uncs_roi are not defined maybe `np.sqrt(self.y_roi + 1)` (never should be zero!)

jvavrek:
> I would say `(y_eval - self.y_roi) / self.y_uncs_roi` or something along these lines, if y_uncs_roi are not defined maybe `np.sqrt(self.y_roi + 1)` (never should be zero!)

This would give errors in sigma, which is useful (maybe a future addition!), but here I just wanted residuals relative to the model. I.e. if the model predicts 100 and the data is 120, then the relative residual is (120-100)/100 = 0.2.

cosama:
@jvavrek You can leave relative residuals if you like, please document it though. The optimization uses residuals normalized by the uncertainties (or sqrt of counts if not provided, that is how we implemented it), so I think relative residuals only tell you so much.

jvavrek:
@cosama I added options for absolute, relative, and sigma residuals. They're documented inline, since I'm less familiar with the rest of the options to write a full docstring.

jccurtis:
> > I would remove the `abs`. I just check and `GaussModel` and `GaussErfModel` both set the parameter bounds to not allow negative values. I think that should be sufficient. Just need to make sure it passes all the tests.
> > Am I the only one who got two figure each time I call the `custom_plot` method in jupyter-notebook?
> 
> It only plots once if you put a semicolon at the end of the line or you call `plt.show()` immediately after.

This depends on the backend. It is standard practice to return the axis in a "simple" plot method. If you return the fig (and maybe the axis) the notebook sees that as an output and visualizes with its internal magic.

jccurtis:
> get this MR in sooner rather than later? Or can we port (some of) them to new issues? (For instance, parameter bounds instead of applying `abs` in the fit functions.)

@jvavrek did you impement this? It would be good to have the default ranges be implemented. I've found that the `abs` in the functions makes things converge faster and I'll (knock on wood) never fit an upside down peak 🤔 . That said, generalization is good and having sane defaults in the param config is key here.

jccurtis:
> @cosama @jccurtis I've made some updates:
> 
>     * an option for relative residuals in the residual plot below the fit plot
> 
>     * a horizontal dashed line at 0 in the residuals plot
> 
>     * in the fit report, print only unsigned uncertainties
> 
>     * updates to the fit model in the fitting notebook so that it runs out of the box
> 
> 
> Here's the example with the relative residuals:
> image
> 
> I also wanted to check if the gauss amplitude guess is being set correctly. It looks like it's being set based on the sum of all counts in the ROI in `GaussModel.guess()`. Why not just the max of the ROI? That should reduce the impact of the background counts.

@jvavrek can you include the new version of this figure?

jccurtis:
> I would say `(y_eval - self.y_roi) / self.y_uncs_roi` or something along these lines, if y_uncs_roi are not defined maybe `np.sqrt(self.y_roi + 1)` (never should be zero!)

Instead of adding one to counts, just clip: `np.sqrt(np.clip(self.y_roi, min=1))`

jvavrek:
@jccurtis 

sigma:

image


relative:

image

jvavrek:
> > get this MR in sooner rather than later? Or can we port (some of) them to new issues? (For instance, parameter bounds instead of applying `abs` in the fit functions.)
> 
> @jvavrek did you impement this? It would be good to have the default ranges be implemented. I've found that the `abs` in the functions makes things converge faster and I'll (knock on wood) never fit an upside down peak 🤔 . That said, generalization is good and having sane defaults in the param config is key here.

@cosama mentioned this was already implemented in `GaussModel`

jccurtis:
> @jccurtis
> 
> sigma:
> 
> image
> 
> relative:
> 
> image

Thanks! Are these with your new guess function?

becquerel/core/fitting.py
```diff
@@ -0,0 +1,830 @@
...
+            if i == 0:
+                label = 'Centroid and FWHM'
+            else:
+                label = None
+            fit_ax.axvline(c, color='#ff7f00')
+            fit_ax.axvspan(c - f / 2.0, c + f / 2.0, color='#ff7f00',
+                           alpha=0.2, label=label)
+        # Misc
+        fit_ax.legend(loc='upper right')
+        # TODO: add ylabel based on units of y
```
jccurtis, line 684: Should this be an issue?

jvavrek:
@jccurtis those are with the old guess function. We can look into improving guesses in the future.

jccurtis:
> > > get this MR in sooner rather than later? Or can we port (some of) them to new issues? (For instance, parameter bounds instead of applying `abs` in the fit functions.)
> > 
> > 
> > @jvavrek did you impement this? It would be good to have the default ranges be implemented. I've found that the `abs` in the functions makes things converge faster and I'll (knock on wood) never fit an upside down peak 🤔 . That said, generalization is good and having sane defaults in the param config is key here.
> 
> @cosama mentioned this was already implemented in `GaussModel`

@jvavrek great thanks. These are hinted in `set_param_hint` in the built in `Model`'s

jccurtis:
@jvavrek thanks for the feedback! I'll take a crack at the docs over the holiday and then I think we can get this in. I'll reach out for clarification for the chunks I didn't write 🤣

becquerel/core/fitting.py
jvavrek: I typically default to 'counts per bin' and update it later with something like
```py
ax.set_ylabel('counts per {0:.1f} keV'.format(binwidths[0]))
```
I guess here we can get the units from `Spectrum`, too. A new issue for this makes sense to me.

jvavrek:
> @jvavrek thanks for the feedback! I'll take a crack at the docs over the holiday and then I think we can get this in. I'll reach out for clarification for the chunks I didn't write 🤣

@jccurtis did you get a chance on the docs? I can try to chip in here, too

becquerel/core/fitting.py
```diff
@@ -0,0 +1,859 @@
...
+        ]
+
+
+class LineModel(Model):
+
+    def __init__(self, *args, **kwargs):
+        super(LineModel, self).__init__(line, *args, **kwargs)
+
+    def guess(self, y, x=None, dx=None, num=2):
+        _, b = _xy_left(y, x=x, num=num)
```
jvavrek, line 102: This `b` is never used

becquerel/core/fitting.py
```diff
@@ -0,0 +1,859 @@
...
+            ('{}amp'.format(self.prefix), 'min', 0.0),
+        ]
+
+
+class ExpGaussModel(Model):
+    """A model of an Exponentially modified Gaussian distribution
+    (see https://en.wikipedia.org/wiki/Exponentially_modified_Gaussian_distribution)
+    """
+
+    fwhm_factor = 2*np.sqrt(2*np.log(2))
```
jvavrek, line 240: This is defined already at the top as `FWHM_SIG_RATIO`

becquerel/core/fitting.py
```diff
@@ -0,0 +1,871 @@
...
+            y_roi_norm = self.y_roi / self.dx_roi
+        if backend.lower().strip() == 'lmfit':
+            weights = self.y_unc_roi ** -1.0
+            self.result = self.model.fit(
+                y_roi_norm, self.params, x=self.x_roi, weights=weights)
+        elif backend.lower().strip() == 'lmfit-pml':
+            self._set_likelihood_residual()
+            self.result = self.model.fit(
+                self.y_roi, self.params, #self.result.params,
+                x=self.x_roi, weights=self.dx_roi,
```
jvavrek, line 573: Are the weights correct here? If so, let's add a note why.

markbandstra:
> @cosama I added options for absolute, relative, and sigma residuals. They're documented inline, since I'm less familiar with the rest of the options to write a full docstring.

From our meeting: could you please add these to be accessible programmatically instead of just inside plotting.

becquerel/core/fitting.py
markbandstra: Leave this alone for now and address in #224

becquerel/core/fitting.py
jccurtis: > @jccurtis those are with the old guess function. We can look into improving guesses in the future.

Will be resolved in #225

becquerel/core/fitting.py
jccurtis: Please do that

becquerel/core/fitting.py
jccurtis: Then this func call this is not needed right?

becquerel/core/fitting.py
jccurtis: - [ ] Add a note that this is converting to counts since the inverse of the weights are applied

becquerel/core/fitting.py
jvavrek: Correct, I'll delete it

becquerel/core/fitting.py
jvavrek: Actually, it's not even called. Deleted!

jvavrek:
From way up above:
> * [x]  Add fitter method to `Spectrum` class

is now in-progress in #218

jvavrek:
Anything else we want done on this one? I think that's everything from today's meeting.

becquerel/core/fitting.py
cosama: We decided to leave this.

markbandstra (approved):
Great work everyone! Time at long last to get this merged.

PR #152, opened 2018-10-25
Remove _examples from example notebooks

jccurtis:
To be consistent ...


jccurtis:
@markbandstra can you check this quickly? It should be a low lift ...

PR #155, opened 2018-11-28
SPE File Parser handle $ENDRECORD and other $key, value fields

### LINKED ISSUE #81: SPE File Parser $ENDRECORD
Whenever I convert a file with PeakEasy to an SPE file it adds a line at the end of the file (with nothing following):

```
$ENDRECORD:
```

This prints a warning:

```
Unknown line:  $ENDRECORD:
```

I think this line is a generic metadata tag to contain other information that cannot be natively saved within a `SPE` file (i.e. when converting a more complex file like `SPC` or `N42`). 

I propose that we:

- [x] Figure out if this line has other specific meaning in an SPE file (besides my above guess)
- [x] Add the data to a `metadata` or other generic text blob in `Spectrum`

markbandstra:
Fix #81 by reading SPE into `metadata` dict.


markbandstra:
This PR adds two new SPE files converted from other formats by @jccurtis in 3673109d

tests/parsers_test.py
```diff
@@ -11,7 +11,7 @@
 SAMPLES_PATH = os.path.join(os.path.dirname(__file__), 'samples')
 SAMPLES = {}
 for extension in ['.spe', '.spc', '.cnf']:
-    filenames = glob.glob(os.path.join(SAMPLES_PATH, '*.*'))
+    filenames = glob.glob(os.path.join(SAMPLES_PATH + '*', '*.*'))
```
jccurtis: @markbandstra was this not catching all files?

markbandstra: I kept the original files in `samples` and put the converted ones in `samples_converted`. We can do this differently but I wanted the provenance to be clear.

PR #173, opened 2019-01-29
Changed two plots to plot cps instead of counts

kcmeehan:
Edited two plots in the overview example notebook to plot cps instead of counts.


jccurtis:
This looks good to me. Any thoughts @cosama @markbandstra ?

jccurtis:
@cosama @markbandstra Thoughts?

PR #187, opened 2019-10-08
Add listmode and bin vs channel definition

### LINKED ISSUE #183: Spectrum should take list-mode data
It'd be nice to create a Spectrum object from list-mode data such as an `np.array()` or from a (possibly time-stamped) plaintext file of ADC values. (See also #116.) This mostly just saves the user the step of calling
`bin_contents, bin_edges = np.histogram(data, nbins, erange)`
before
`bq.Spectrum(counts=bin_contents)`
but it should be integrated into Becquerel.

### LINKED ISSUE #184: energy calibrations should use channel space, not bin space
`becquerel`  classes should work with an x-axis of channels, not bins. For example, say I have an ADC that produces a spectrum of 32768 discrete channels, but I need to rebin the spectrum by a factor of 4 to 8192 channels to trade off resolution for statistics. `becquerel` would give energy calibration gains differing by a factor of 4 for the original and rebinned spectra, because it works in the (implicit) bin space of the `spectrum` rather than the (explicit) channel space.

This is even more crucial for list-mode data (#183) with an arbitrarily large number of possible ADC values. The user may need to play around with different bin widths, and the calibration should not vary as a result.

Working in channel space would also allow the user to apply the calibration from a calibration spectrum to a production spectrum with different binning.

My current hotfix is to add a `self.binwidth` member to `autocal` and then do
```
fit = find_best_gain(self.peakfinder.channels * self.binwidth, ...)
```
because `peakfinder.channels` really returns bins, not channels, where it thinks peaks are located. This leads to difficulties down the road with plotting calibration curves, e.g., as the `peakfinder.channels` are still in bin space, not channel space. A more robust solution should therefore probably pass the spectrum bin width to the `peakfinder` and/or `spectrum` classes instead, so that by the time `autocal` is used, it is already seeing channels and not bins.

Note that working in channel space instead of bin space will not break any existing examples (e.g. `examples/autocal.ipynb`) because the binwidths there are just 1 channel/bin.

jvavrek:
This is a working solution to read in list-mode data, based on @jccurtis's suggestion in #183 . It also includes additions to the `spectrum.ipynb` notebook and further tests in `spectrum_test.py`. Feel free to re-format the latter to make it more cohesive with the rest of the tests.

This issue is coupled with #184 , as working with list mode data can be tricky due to the channel vs bin issue.

--- 

Closes #183
Closes #184


jccurtis:
@jvavrek nice work! Thanks for this. Just a quick note to add a `warning` in the `classmethod` since `bin_edges_keV` may not be accurate ...

becquerel/core/spectrum.py
```diff
@@ -456,6 +461,62 @@ def from_file(cls, infilename):
...
+          xmax: maximum x of histogram; equals bin_edges[-1] + bin_widths[-1]
+          timestamps: array of timestamps associated with listmode_data
+
+        Returns:
+          A Spectrum object
+
+        Raises:
+          AssertionError: no listmode_data, or xmin >= xmax, or nbins < 1
+        """
+
```
jccurtis: Add a warning to this method since `bin_edges_keV` could be misleading.

becquerel/core/spectrum.py
```diff
@@ -456,6 +461,62 @@ def from_file(cls, infilename):
...
+            assert len(timestamps) == len(listmode_data)
+            # TODO: How do we handle these? This isn't accurate ...
+            # What are the timestamps: 'float', 'int',
+            #   'datetime', epoch [nano][micro][milli]seconds?
+            warnings.warn('Using first/last timestamps to approximate ' +
+                          'start / stop_time and realtime.', SpectrumWarning)
+            kwargs['start_time'] = min(timestamps)
+            kwargs['stop_time'] = max(timestamps)
+            kwargs['realtime'] = kwargs['stop_time'] - kwargs['start_time']
+            kwargs['livetime'] = None
```
cosama: I think it would be more intuitive if instead of setting those from a timestamp array that can change from use case to use case, we just pass all the unused kwargs to the class creator. So if someone cares about `livetime`, he can set it the way he thinks it needs to be set, i.e. `livetime=timstamps[-1] - timestamps[0]`, (same for `start_time`, `stop_time` and `realtime`, which are never really used anyways). We can not impossible cover all cases that might arise here, better let the user make a wise decision.

becquerel/core/spectrum.py
```diff
@@ -726,6 +787,72 @@ def downsample(self, f, handle_livetime=None):
...
+        """
+        assert x >= self.bin_edges_kev[0]
+        assert x < self.bin_edges_kev[-1]
+
+        if self.has_uniform_bins():
+            return int((x - self.bin_edges_kev[0]) / self.bin_widths[0])
+        else:
+            return max(bisect_left(self.bin_edges_kev, x) - 1, 0)
+
+    def get_nbins(self):
```
cosama: why not define `__len__` instead

becquerel/core/spectrum.py
```diff
@@ -726,6 +787,72 @@ def downsample(self, f, handle_livetime=None):
...
+        Returns:
+          The integer bin number that contains x
+        """
+        assert x >= self.bin_edges_kev[0]
+        assert x < self.bin_edges_kev[-1]
+
+        if self.has_uniform_bins():
+            return int((x - self.bin_edges_kev[0]) / self.bin_widths[0])
+        else:
+            return max(bisect_left(self.bin_edges_kev, x) - 1, 0)
```
cosama: why not use `numpy.searchsorted(self.bin_edges_kev, x, side='left')`, wouldn't add another dependency, and is probably the fastest in any case. I can't imagine a case where `bin_edges_kev` wouldn't be sorted, and if they are not, we probably should enforce it on creation.

cosama: Another comment. Also we probably should figure out an answer to all the FIXME's. Didn't look to much into those.

jvavrek: Thanks, I didn't know about the numpy solution. I'll also update it to use the `'right'` option, as that makes more sense for an inclusive low edge but exclusive high edge.

cosama: Go for it, I had the `left` because it was the default, haven't thought much about what makes more sense.

becquerel/core/spectrum.py
cosama: Just checked, the __len__ method is already defined, so `len(spectrum)` should give you the right answer here. I don't think this adds any new abilities.

jccurtis: Agreed, good catch @cosama

jvavrek: Agreed, removed

becquerel/core/spectrum.py
jvavrek: Sure, makes sense. This also saves us from converting back and forth between `datetime` and `float` etc depending on timestamps vs time durations like `livetime` and `realtime`, and saves us from always approximating the start/stop acquisition times as the first/last timestamp.

Since the `kwargs` are already passed to the class creator, this just becomes:

    ...
    if timestamps is not None:
            assert len(timestamps) == len(listmode_data)
    
    return cls(**kwargs)

cosama: I would not have a timestamps field at all as it won't be used. Something like

```
    def from_listmode(cls, listmode_data, timestamps=None, bins=None,
                      xmin=None, xmax=None, **kwargs):
        ...
        kwargs['counts'] =  bin_counts
        kwargs['bin_edges_kev' = bin_edges
        ...
        return cls(**kwargs)
```

jvavrek: Fine by me. I've removed them in the latest commit. The user can always do `spec.timestamps = ...` if necessary.

jvavrek:
Is this ready to merge?

jvavrek:
The latest three commits implement `bin_edges_raw` from #184 in order to correctly build a `Spectrum` object from raw list mode data, and also implement a few new necessary methods as a result. Because we keep both the `raw` and `kev` x-axes, there is a fair bit of almost-duplicate code, such as the properties `bin_centers_raw` and `bin_centers_kev`. I think that is a reasonable amount of duplication, but for the more complicated methods `find_bin()` and `has_uniform_bins()`, I opted instead for passing a boolean `use_kev` argument. I don't particularly like this solution, so if anyone has a more elegant way, please let me know.

All tests pass, but I suspect some of the other notebooks will break slightly as a result.

jccurtis:
@jvavrek added the issues (#183, #184) which will be resolved by this PR. Please add others if needed. Almost there!

becquerel/core/spectrum.py
```diff
@@ -341,24 +351,72 @@ def cpskev_vals(self):
...
+        Returns:
+          np.array of int's from 0 to (len(self.counts) - 1)
+        """
+
+        return np.arange(len(self), dtype=int)
+
     @property
     def channels(self):
-        """Channel index.
+        """Bin indexes.
```
markbandstra: Indices? :)

jvavrek: Yeah, good point

jvavrek:
I'm still updating some documentation, so hopefully this should be ready to go soon.

becquerel/core/spectrum.py
```diff
@@ -370,11 +428,30 @@ def energies_kev(self):
...
+        warnings.warn('energies_kev is deprecated and will be removed in a '
+                      'future release. Use bin_centers_kev instead.')
+
         if not self.is_calibrated:
             raise UncalibratedError('Spectrum is not calibrated')
         else:
             return bin_centers_from_edges(self.bin_edges_kev)
 
+    @property
+    def bin_widths_kev(self):
```
markbandstra: Good call on this method and `bin_widths_raw`

becquerel/core/spectrum.py
```diff
@@ -386,6 +463,9 @@ def bin_widths(self):
           UncalibratedError: if spectrum is not calibrated
         """
 
+        warnings.warn('bin_widths is deprecated and will be removed in a '
```
markbandstra: Can you raise a [`DeprecationWarning`](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) if someone wants to catch this type of warning.

becquerel/core/spectrum.py
```diff
@@ -722,9 +894,81 @@ def downsample(self, f, handle_livetime=None):
...
+        Args:
+          x: value whose bin to find
+          use_kev: check bin_edges_kev if True, otherwise bin_edges_raw
+
+        Raises:
+          AssertionError: if x is outside the bin edges or equal to up edge
+          SpectrumError: if use_kev=True but Spectrum is not calibrated
+
+        Returns:
+          The integer bin number that contains x
```
markbandstra: There's a subtle semantic issue with this method, and I would suggest renaming to `find_bin_index` or something of the sort. The term "bin" is a little ambiguous as to whether it is referring to the index of the bin or, e.g., the value of the bin center. Do others have the same confusion?

jvavrek:
For now I have decided against changing every instance of `channel` in `autocal.py` and `energycal.py`, since that kind of change would propagate to methods such as `ch2kev()` and `kev2ch()`. I've left a note to this effect in the class constructors. We could however go with `raw2kev()` etc.  Thoughts?

becquerel/core/spectrum.py
jvavrek: Sure, but it looks like the output is suppressed by default in that case.

becquerel/core/spectrum.py
jvavrek: I'm used to the ROOT convention where a "bin" is always an integer index, and methods exist to compute bin low/up edges or centers. See, e.g., [here](https://root.cern.ch/doc/master/classTH1.html#ae0895b66e993b9e9ec728d828b9336fe).

becquerel/core/spectrum.py
cosama: I have seen that before, I think we want them being shown to everybody, so we should turn them on. Use:

```
warnings.simplefilter('always', DeprecationWarning)
```

In the header right after the import, we should fix that in other places too.

cosama: It looks like there is also an entry in `setup.cfg` to ignore deprecation warnings:

```
setup.cfg:    ignore:.*getargspec.*:DeprecationWarning
```

Does this only affect submodules or this module as well?

jvavrek: I'm not familiar with how the `setup.cfg` file works, but adding the `simplefilter` line above works well. Let me know if I should remove the line from `setup.cfg` too.

jccurtis:
The battle of `channel` vs `bin` rages on!!

becquerel/core/spectrum.py
```diff
@@ -12,6 +12,8 @@
 from . import plotting
 import warnings
 
+EPS = np.finfo(float).eps
```
jccurtis: This should be in `utils`?

becquerel/core/spectrum.py
```diff
@@ -341,24 +353,73 @@ def cpskev_vals(self):
...
+        """Uncertainties on the counts per second per keV in each bin.
 
         Returns:
           an np.array of floats
         """
 
         return unumpy.std_devs(self.cpskev)
 
+    @property
+    def bins(self):
```
markbandstra: Please change to `bin_indices`

becquerel/core/spectrum.py
```diff
@@ -57,19 +61,24 @@ class Spectrum(object):
...
+      counts_vals: counts in each bin, no uncertainty
+      counts_uncs: uncertainties on counts in each bin
+      cps: counts per second in each bin, with uncertainty
+      cps_vals: counts per second in each bin, no uncertainty
+      cps_uncs: uncertainties on counts per second in each bin
+      cpskev: CPS/keV in each bin, with uncertainty
+      cpskev_vals: CPS/keV in each bin, no uncertainty
+      cpskev_uncs: uncertainties on CPS/keV in each bin
+      bins: np.array of bin index as integers
+      channels: np.array of bin index as integers (deprecated)
```
jccurtis: Add a note that this warns.

becquerel/core/spectrum.py
```diff
@@ -341,24 +353,73 @@ def cpskev_vals(self):
...
+
+        return np.arange(len(self), dtype=int)
+
     @property
     def channels(self):
-        """Channel index.
+        """Bin indices.
 
         Returns:
           np.array of int's from 0 to (len(self.counts) - 1)
```
jccurtis: This is an alias for `self.bins`. Add that this warns and will be removed in a future version.

becquerel/core/spectrum.py
```diff
@@ -456,6 +572,52 @@ def from_file(cls, infilename):
...
+                      **kwargs):
+        """Construct a Spectrum object (specifically the `bin_edges_raw` and
+        `counts` of a histogram) from an array of listmode data. It is left to
+        the user to set kwargs realtime, livetime, etc, rather than trying to
+        cover all use cases here.
+
+        Args:
+          listmode_data: the array containing the listmode data, e.g., the ADC
+            values triggered or pulse integral values recorded
+          bins: integer number of bins OR array of bin_edges
```
jccurtis: Add that this is numpy.histogram style.

becquerel/core/spectrum.py
jccurtis: This has been resolved.

jvavrek: Now called `find_bin_index()`

becquerel/core/spectrum.py
jvavrek: Moved to `utils`

jvavrek:
That's the last of the post-meeting comments. Ready to merge?

becquerel/core/spectrum.py
```diff
@@ -7,11 +7,13 @@
 import numpy as np
 from uncertainties import UFloat, unumpy
 from .. import parsers
-from .utils import handle_uncs, handle_datetime, bin_centers_from_edges
+from .utils import handle_uncs, handle_datetime, bin_centers_from_edges, EPS
 from .rebin import rebin
 from . import plotting
 import warnings
 
+warnings.simplefilter('default', DeprecationWarning)
```
jccurtis: @jvavrek move to __init__ instead of repeating

becquerel/core/rebin.py
```diff
@@ -2,6 +2,7 @@
 import numpy as np
 import numba as nb
 import warnings
+warnings.simplefilter('always', DeprecationWarning)
```
jccurtis: @jvavrek move to __init__ instead of repeating

becquerel/core/peakfinder.py
```diff
@@ -5,8 +5,11 @@
 from __future__ import print_function
 import matplotlib.pyplot as plt
 import numpy as np
+import warnings
 from .spectrum import Spectrum
 
+warnings.simplefilter('always', DeprecationWarning)
```
jccurtis: @jvavrek move to __init__ instead of repeating

becquerel/parsers/spectrum_file.py
```diff
@@ -5,6 +5,7 @@
 import warnings
 import numpy as np
 from scipy.interpolate import interp1d
+warnings.simplefilter('always', DeprecationWarning)
```
jccurtis: @jvavrek move to __init__ instead of repeating

becquerel/parsers/spe_file.py
```diff
@@ -8,6 +8,7 @@
 import numpy as np
 from .spectrum_file import (SpectrumFile, SpectrumFileParsingError,
                             SpectrumFileParsingWarning)
+warnings.simplefilter('always', DeprecationWarning)
```
jccurtis: @jvavrek move to __init__ instead of repeating

examples/misc.ipynb
```diff
@@ -674,7 +674,7 @@
      "name": "stderr",
      "output_type": "stream",
      "text": [
-      "becquerel/core/spectrum.py:556: SpectrumWarning: Subtraction of counts-based specta, spectra have been converted to CPS\n",
+      "/Users/jvavrek/becquerel/becquerel/core/spectrum.py:713: SpectrumWarning: Subtraction of counts-based specta, spectra have been converted to CPS\n",
```
jccurtis: Remove your username here

examples/energycal.ipynb
```diff
@@ -41,8 +41,8 @@
      "name": "stderr",
      "output_type": "stream",
      "text": [
-      "becquerel/parsers/spectrum_file.py:122: UserWarning: Ignoring calibration; energies not monotonically increasing\n",
-      "  'Ignoring calibration; energies not monotonically increasing')\n"
+      "/Users/jvavrek/becquerel/becquerel/parsers/spectrum_file.py:135: SpectrumFileParsingWarning: Spectrum will be initated without an energy calibration;invalid calibration, energies not monotonically increasing.\n",
```
jccurtis: Remove username from path

tests/spectrum_test.py
```diff
@@ -228,6 +227,52 @@ def test_negative_input(spec_data):
     assert np.any(np.isnan(spec.counts_uncs))
 
 
+# ----------------------------------------------
+#      Test Spectrum.from_listmode behavior
+# ----------------------------------------------
+
+# It's easy to introduce off-by-one errors in histogramming, so run quite a few
```
jccurtis: Move these comments to the `docstring`

examples/misc.ipynb
jvavrek: That's just the output from a warning in the notebook

examples/energycal.ipynb
jvavrek: Again, just a warning output

jccurtis:
@jvavrek sorry I just added a bunch of extra comments.

jccurtis:
This is ready for merging. Nice work @jvavrek!

PR #188, opened 2019-10-13
Feature ecal models #99

### LINKED ISSUE #99: Additional energy calibration models
Building off of #33, #41, we should add additional energy calibration models using `EnergyCalBase`

jccurtis:
I thought we had already taken care of this. I need to add tests and another model but afterwards this should be ready to go! I plan to knock this out in Nov.

Closes #99


becquerel/core/energycal.py
```diff
@@ -390,3 +390,47 @@ def _perform_fit(self):
...
+    def _kev2ch(self, kev):
+        """Convert energy value(s) to channel(s).
+        Args:
+          kev: an np.array, float, or int of energy values [keV]
+        Returns:
+          the channel value(s) corresponding to the input energies.
+            a float if input is scalar. an np.array if input is iterable
+        """
+        # TODO: address with the quadratic inverse formula
+        raise NotImplementedError('Sorry')
```
jvavrek: This could be implemented. A unique inverse for all `ch` <-> `kev` pairs exists if the `ch` domain only contains one side of the parabola, e.g. `ch[0] > -b/(2*a)`. `ValueError` if the `ch` does not have a unique `kev`.

PR #193, opened 2019-12-12
Fix Iterable import.

jccurtis:
This import will be removed in py38 and cause an ImportError:

```
DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
```


jccurtis:
@markbandstra bumping this

jccurtis:
Rerunning tests after dev merge ... then I'll finish this

PR #196, opened 2019-12-17
Release 0.2.4

jccurtis:


jccurtis:
This release bump is due to some recent PR's which didn't make it into the release earlier today.

markbandstra:
@jccurtis please add a clarification to the first steps of `RELEASING.txt` so that it is clear that we need to pull from the latest `develop` or `master` so that we can avoid having to do this in the future.

jccurtis:
@markbandstra done!

PR #199, opened 2020-02-11
let find_bin_index find indices for multiple x values

### LINKED ISSUE #198: Spectrum `find_bin_index` should be able to accept multiple x-values
Spectrum `find_bin_index` should be able to accept multiple x-values and return their bin indices. This should be pretty simple to implement.

jvavrek:
Fixes #198, including updated documentation, tests, and notebooks.

### DIFF
--- a/becquerel/core/spectrum.py
+++ b/becquerel/core/spectrum.py
@@ -973,7 +973,7 @@ class Spectrum(object):
 
         bin_edges = self.bin_edges_kev if use_kev else self.bin_edges_raw
         bin_widths = self.bin_widths_kev if use_kev else self.bin_widths_raw
-        x = np.array(x)
+        x = np.asarray(x)
 
         if np.any(x < bin_edges[0]):
             raise SpectrumError('requested x is < lowest bin edge')
--- a/becquerel/core/spectrum.py
+++ b/becquerel/core/spectrum.py
@@ -976,9 +976,9 @@ class Spectrum(object):
         x = np.array(x)
 
         if np.any(x < bin_edges[0]):
-            raise ValueError('requested x is < lowest bin edge')
+            raise SpectrumError('requested x is < lowest bin edge')
         if np.any(x >= bin_edges[-1]):
-            raise ValueError('requested x is >= highest bin edge')
+            raise SpectrumError('requested x is >= highest bin edge')
 
         if self.has_uniform_bins(use_kev=use_kev):
             return ((x - bin_edges[0]) / bin_widths[0]).astype(np.int)
--- a/becquerel/core/spectrum.py
+++ b/becquerel/core/spectrum.py
@@ -960,8 +960,8 @@ class Spectrum(object):
           use_kev: check bin_edges_kev if True, otherwise bin_edges_raw
 
         Raises:
-          ValueError: if x is outside the bin edges or equal to up edge
-          SpectrumError: if use_kev=True but Spectrum is not calibrated
+          SpectrumError: if use_kev=True but Spectrum is not calibrated; or if
+            x is outside the bin edges or equal to up edge
 
         Returns:
           The integer bin index or indices containing x
--- a/becquerel/core/spectrum.py
+++ b/becquerel/core/spectrum.py
@@ -948,7 +948,7 @@ class Spectrum(object):
         return True
 
     def find_bin_index(self, x, use_kev=True):
-        """Find the Spectrum bin index that contains an x-axis value x.
+        """Find the Spectrum bin index or indices containing x-axis value(s) x.
 
         If the Spectrum has uniform binning, then we just solve the linear
         equation between bin indices and x-axis values. Otherwise, we use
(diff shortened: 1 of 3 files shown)


becquerel/core/spectrum.py
```diff
@@ -973,12 +973,15 @@ def find_bin_index(self, x, use_kev=True):
 
         bin_edges = self.bin_edges_kev if use_kev else self.bin_edges_raw
         bin_widths = self.bin_widths_kev if use_kev else self.bin_widths_raw
+        x = np.array(x)
 
-        assert x >= bin_edges[0]
-        assert x < bin_edges[-1]
+        if np.any(x < bin_edges[0]):
+            raise ValueError('requested x is < lowest bin edge')
```
jccurtis: use SpectrumError?

becquerel/core/spectrum.py
```diff
@@ -973,12 +973,15 @@ def find_bin_index(self, x, use_kev=True):
...
+        if np.any(x >= bin_edges[-1]):
+            raise ValueError('requested x is >= highest bin edge')
 
         if self.has_uniform_bins(use_kev=use_kev):
-            return int((x - bin_edges[0]) / bin_widths[0])
+            return ((x - bin_edges[0]) / bin_widths[0]).astype(np.int)
```
jccurtis: this changes the output type for scalar input. I think this should return scalar output for scalar input (as the docstring implies). See:

https://docs.scipy.org/doc/numpy/reference/generated/numpy.isscalar.html

You could not do the conversion (the edge checks still work) and then handle the dtype based on `isscalar`

examples/spectrum.ipynb
```diff
@@ -186,7 +177,7 @@
      "name": "stderr",
      "output_type": "stream",
      "text": [
-      "becquerel/core/spectrum.py:440: DeprecationWarning: energies_kev is deprecated and will be removed in a future release. Use bin_centers_kev instead.\n",
+      "/Users/jvavrek/becquerel/becquerel/core/spectrum.py:440: DeprecationWarning: energies_kev is deprecated and will be removed in a future release. Use bin_centers_kev instead.\n",
```
jccurtis: home ... we should really write a clean script for this

examples/spectrum.ipynb
```diff
@@ -741,19 +732,23 @@
...
-    "print(spec7.find_bin_index(9999, use_kev=False))"
+    "print(spec7.find_bin_index(9999, use_kev=False))\n",
+    "print(spec7.find_bin_index(np.array([4000, 5000, 6000]), use_kev=False))\n",
+    "print(spec8.find_bin_index(np.array([4000, 5000, 6000]), use_kev=False))"
    ]
   },
   {
    "cell_type": "code",
-   "execution_count": 22,
+   "execution_count": 27,
```
jccurtis: executed out of order? please run from scratch 😄

becquerel/core/spectrum.py
jvavrek: There are a few other places in `spectrum.py` where a `ValueError` is raised rather than a `SpectrumError`, so I'm not sure what's the most consistent. I'll go with the latter for now.

becquerel/core/spectrum.py
jvavrek: As we found offline, this (inadvertently) takes advantage of some subtle [numpy casting rules](https://docs.scipy.org/doc/numpy/reference/generated/numpy.can_cast.html). With the `x = np.array(x)` above these lines, a scalar input `x` will still return an `np.int` rather than an array. This is why the tests passed and the notebooks looked correct. I may add some more type-checking to the tests to be safe.

jccurtis: Yeah a zero dim array with int dtype behaves like an int. Good stuff m!

becquerel/core/spectrum.py
jccurtis: Yeah, I think those are all legacy...

becquerel/core/spectrum.py
```diff
@@ -973,12 +973,15 @@ def find_bin_index(self, x, use_kev=True):
 
         bin_edges = self.bin_edges_kev if use_kev else self.bin_edges_raw
         bin_widths = self.bin_widths_kev if use_kev else self.bin_widths_raw
+        x = np.array(x)
```
markbandstra: Maybe `np.asarray` would be a little clearer and we do not need a separate copy of `x`.

tests/spectrum_test.py
```diff
@@ -253,9 +253,13 @@ def test_listmode():
     assert spec0.find_bin_index(XMIN, use_kev=False) == 0
     assert spec0.find_bin_index(XMIN + BW/4.0, use_kev=False) == 0
     assert spec0.find_bin_index(XMAX - BW/4.0, use_kev=False) == NBINS - 1
-    with pytest.raises(AssertionError):
+    assert np.all(spec0.find_bin_index(
+        spec0.bin_edges_raw[:-1], use_kev=False) == np.arange(NBINS))
```
markbandstra: Let's add tests that use energies and `use_kev=True`

Issue #200, opened 2020-02-12
Split up test_listmode


jvavrek:
Split `test_listmode` in `spectrum_test.py` into separate functions focusing on listmode vs `find_bin_index`. Can use fixtures for this. Also add tests with `use_kev=True`.

jvavrek:
This was closed via #201

PR #201, opened 2020-03-02
Issue tests #200: Spectrum tests and associated changes to bin-related methods and docstrings

jvavrek:
This splits up a large block of tests I had written previously into several smaller tests focusing separately on list mode and find index functionalities.

### DIFF
--- a/becquerel/core/spectrum.py
+++ b/becquerel/core/spectrum.py
@@ -959,10 +959,12 @@ class Spectrum(object):
     def find_bin_index(self, x, use_kev=None):
         """Find the Spectrum bin index or indices containing x-axis value(s) x.
 
-        If the Spectrum has uniform binning, then we just solve the linear
-        equation between bin indices and x-axis values. Otherwise, we use
-        searchsorted to bisect for the insertion point where x would fit in a
-        list of bin edges, then subtract 1 to get the index of the low edge.
+        One might think that if the Spectrum has uniform binning, we could just
+        solve the linear equation between bin indices and x-axis values.
+        However, this can introduce enough precision loss to change the index.
+        As such, we use searchsorted to bisect for the insertion point where x
+        would fit in a list of bin edges, then subtract 1 to get the index of
+        the low edge. The numpy searchsorted method is only ~1.4 us per loop.
 
         Args:
           x: value(s) whose bin to find
@@ -992,10 +994,7 @@ class Spectrum(object):
         if np.any(x >= bin_edges[-1]):
             raise SpectrumError('requested x is >= highest bin edge')
 
-        if self.has_uniform_bins(use_kev=use_kev):
-            return ((x - bin_edges[0]) / bin_widths[0]).astype(np.int)
-        else:
-            return np.searchsorted(bin_edges, x, 'right') - 1
+        return np.searchsorted(bin_edges, x, 'right') - 1
 
     def get_bin_properties(self, use_kev=None):
         """Convenience function to get bin properties: edges, widths, centers.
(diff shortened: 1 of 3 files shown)


jvavrek:
TODO:
- [ ] update tests once #203 is merged

tests/spectrum_test.py
```diff
@@ -231,58 +231,115 @@ def test_negative_input(spec_data):
...
-    # additional type checking
+    assert len(spec) == NBINS
+    assert np.isclose(widths[0], BW)
+    assert edges[0] == XMIN
+    assert edges[-1] == XMAX
+    assert len(edges) == NBINS + 1
+    assert spec.has_uniform_bins(use_kev=use_kev)
+
+
+@pytest.mark.parametrize('spec', [
```
markbandstra: Test calibrated and uncalibrated spectra with different binnings

tests/spectrum_test.py
```diff
@@ -231,58 +231,115 @@ def test_negative_input(spec_data):
...
+        spec.find_bin_index([XMIN], use_kev=False), np.ndarray)
+
+
+@pytest.mark.parametrize('spec', [
+    make_spec_listmode('uniform'),
+    make_spec_listmode('log'),
+    make_spec('uncal'),
+    make_spec('cal')])
+def test_index_out_of_bounds(spec):
+    '''Raise a SpectrumError when we look for a bin index out of bounds.'''
```
markbandstra: PEP 257 :)

tests/spectrum_test.py
jvavrek: It looks like `has_uniform_bins()` doesn't always work with calibrated spectra—it tests if all the widths are the same to within `EPS = np.finfo(float).eps` (`2.220e-16` on my machine) but the calibration multiplication can introduce slightly higher deviations.

Acceptable to test against, say, `2*EPS` ?

jvavrek: Actually, more like `100*EPS` might be required...

jvavrek:
I think this is ready to go

jccurtis:
did you update the tests based on #203 as noted above?

becquerel/core/spectrum.py
```diff
@@ -929,51 +929,62 @@ def has_uniform_bins(self, use_kev=True):
...
-                                'uncalibrated Spectrum.')
+            raise UncalibratedError('Cannot access energy bins with an ' +
+                                    'uncalibrated Spectrum.')
 
         bin_widths = self.bin_widths_kev if use_kev else self.bin_widths_raw
         iterator = iter(bin_widths)
         x0 = next(iterator, None)
         for x in iterator:
-            if abs(x/x0 - 1.0) > EPS:
+            if abs(x/x0 - 1.0) > rtol:
```
jccurtis: why not `np.allclose(x, x[0], rtol=rtol)`?

jccurtis:
Nit: update title to reflect changes to `Spectrum`

becquerel/core/spectrum.py
jvavrek: I found this nice solution [here](https://github.com/0cjs/py-allthesame#8-writing-the-code-the-function), but if you prefer the `np` solution for readability I can change it. The current version has the advantage of terminating early but I can't imagine that mattering much in an absolute sense.

jvavrek: Quick tests indicate that the `np` solution is substantially faster for all but the shortest of actually-uniform arrays, starting around 70 elements, while the `iter` solution is substantially faster for most non-uniform arrays such as log bins. Either way, everything is < 100 us.

jvavrek: added a note in the latest commit

PR #205, opened 2020-06-17
Update requirements to deal with python2/3 llvmlite issue

markbandstra:
Fix the issue coming up in Travis with python3 versions:
```
error: llvmlite 0.31.0 is installed but llvmlite<0.34,>=0.33.0.dev0 is required by {'numba'}
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,8 @@ html5lib
 lxml
 matplotlib
 numba>=0.47.0
-llvmlite==0.31.0
+llvmlite==0.31.0; python_version < "3"
+llvmlite; python_version >= "3"
 numpy
 pandas
 python-dateutil
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,6 @@ lxml
 matplotlib
 numba>=0.47.0
 llvmlite==0.31.0; python_version < "3"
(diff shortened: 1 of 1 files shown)


jccurtis:
Nice one guys. The `numba` developers made a mistake with pushing a `llvmlite` version (`0.32.0`) with an invalid python2 tag even though it wasn't python2 compatible. That lead to this hardcoded value. Ultimately `python2` needs to go away ...

markbandstra:
For the record, this uses dependency specification from [PEP 508](https://www.python.org/dev/peps/pep-0508/#environment-markers)

markbandstra:
Just realized this is also a problem on master -- should we do a hotfix?

requirements.txt
```diff
@@ -4,7 +4,8 @@ html5lib
 lxml
 matplotlib
 numba>=0.47.0
-llvmlite==0.31.0
+llvmlite==0.31.0; python_version < "3"
+llvmlite; python_version >= "3"
```
cosama: This line (8) is not required.

markbandstra:
Scratch that - master should be fine with python3 because `requirements.txt` lists `numba` and it doesn't have the `llvmlite` version constraint.

PR #210, opened 2020-09-30
Adjust for recent changes to NNDC API

markbandstra:
It appears that two things have changed on NNDC in recent months:

(1) The error message that is returned when requesting too many wallet cards has changed
(2) The decay mode kwarg no longer seems to be honored in the query

Change (1) was leading to the recent test failures on Travis. Fixing (1) leads to (2) causing a separate test to fail, one that requests all wallet cards for spontaneous fission isotopes (`decay="SF"`), which used to be a small enough number to not matter but now thousands of isotopes are returned. This PR skips that test and raises a warning when specifying the `decay` kwarg.

Other tests whose queries are affected by (2) do not raise errors because they are generally not strict in their testing; they usually check that no error was raised by the query and that a nonzero number of elements was returned. We might want to consider making the NNDC testing more strict, but that could also make our code harder to maintain if their database changes.

### DIFF
--- a/becquerel/tools/nndc.py
+++ b/becquerel/tools/nndc.py
@@ -700,7 +700,9 @@ A  	Element	Z  	N  	Energy  	JPi           	Mass Exc  	Unc  	T1/2 (txt)         
                 raise NNDCInputError(
                     'Decay mode must be one of {}, not {}'.format(
                         WALLET_DECAY_MODE.keys(), kwargs['decay'].lower()))
-            warnings.warn('query kwarg "decay" may not be working on NNDC')
+            warnings.warn('query kwarg "decay" may not be working on NNDC, ' +
+                          'and the user is advised to check the ' +
+                          '"Decay Mode" column of the resulting DataFrame')
             self._data['dmed'] = 'enabled'
             self._data['dmn'] = WALLET_DECAY_MODE[kwargs['decay'].lower()]
         # handle energy level condition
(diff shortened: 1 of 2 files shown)


cosama:
I think we can merge this @jccurtis. Should we do this as a hotfix (i.e. we have to change this to master), or do a normal release?

cosama (approved):

becquerel/tools/nndc.py
```diff
@@ -699,6 +700,7 @@ def update(self, **kwargs):
                 raise NNDCInputError(
                     'Decay mode must be one of {}, not {}'.format(
                         WALLET_DECAY_MODE.keys(), kwargs['decay'].lower()))
+            warnings.warn('query kwarg "decay" may not be working on NNDC')
```
jccurtis, line 703: Do we have a NNDCWarning?

jccurtis:
> I think we can merge this @jccurtis. Should we do this as a hotfix (i.e. we have to change this to master), or do a normal release?

@cosama lets run a new release

becquerel/tools/nndc.py
cosama: No just a `NNDCError`. I personally think custom warnings are not that useful. Haven't really figured out a way how to make use of them. For example it seems to not be possible to catch a certain warning type only.

jccurtis: Okay, yeah I agree. Let's include the `kwargs` in this:

```suggestion
            warnings.warn(
                'query kwarg "decay" may not be working on NNDC, kwargs: {}'
                .format(kwargs)
            )
```

jccurtis: Do we want this suggestion? @markbandstra @cosama ?

markbandstra: It's fine but I don't really see why that extra information is needed. The point is just to flag when a user sets kwarg `"decay"` and notify that the query may work but the behavior probably won't be what they expect, and it's not an issue on our side. The workaround is to select the relevant rows of the returned `DataFrame`. Maybe I'm missing something.

jccurtis: Okay, can the warning be expanded to include your comments then? Thanks!

PR #211, opened 2020-10-02
rebin.py: use nb.guvectorize and allow xarray

chunhochow:
rebin.py:
change from nb.jit to a more exact implementation of nb.guvectorize
other changes to allow the use of xarray DataArrays with rebin.py
new import requirement of xarray


chunhochow:
I don't know how to check what test failed.... Can any of you teach me how to do it? Thanks

cosama:
Click on 'Details' in the column saying 'continuous-integration/travis-ci/pr — The Travis CI build failed'. Then click on any of the red columns and you get the build log.

It doesn't work because it hasn't installed `xarrays`. You need to add it to requirements.txt.

chunhochow:
Thanks @cosama ! The having to click on the red boxes part was not intuitive haha.

jvavrek:
Now it's failing due to #210 ...

becquerel/core/rebin.py
```diff
@@ -100,40 +101,32 @@ def _check_any_overlap(in_edges, out_edges):
...
+#     arr1_shape = arr1.shape
+#     if edges:
+#         if len(arr1_shape) == 1:
+#             arr1_shape = (arr1_shape[0] - 1,)
+#         else:
+#             arr1_shape = (arr1_shape[0], arr1_shape[1] - 1,)
+#     if arr0_shape != arr1_shape:
+#         raise RebinError(
+#             '{}{} does not have a shape compatible with {}{}'.format(
+#                 arr0_name, arr0_shape, arr1_name, arr1_shape))
```
jvavrek, line 126: Maybe just delete entirely, instead of commenting out?

chunhochow: Where's the 😂 emoji. On a more serious note, I'm cool with that too.

cosama:
@chunhochow Have a look if that works for you. I removed some of the checks and added some additional flexibility to out_edges, which can also be broadcast if requested now. Also `xarray` is now a kwarg to `rebin`. Reduces code repetition.

@jccurtis I think this is a nice improvement to the module.  It reduces the amount of code considerably but even increases the flexibility of the method. We can now input any dimensional array as long as the last dimensions are correct, the remaining dimensions are just broadcast.

chunhochow:
@cosama love your idea of putting `import xarray` under control of a kwarg!

becquerel/core/rebin.py
```diff
@@ -70,28 +53,28 @@ def _check_partial_overlap(in_edges, out_edges):
...
             'zeros will padded on the right side of the new spectrum',
             RebinWarning)
 
 
 def _check_any_overlap(in_edges, out_edges):
-    """
+    """In and out edges overlapping at all.
+
     Args:
         in_edges (np.ndarray): an array of the input bin edges (1D or 2D)
```
jccurtis, line 72: Does this still apply (the 1D/2D bin edges)?

becquerel/core/rebin.py
```diff
@@ -394,63 +303,103 @@ def rebin(in_spectra, in_edges, out_edges, method="interpolation",
...
-    if method == "interpolation":
-        if in_spectra.ndim == 2:
-            return _rebin2d_interpolation(
-                in_spectra, in_edges, out_edges, slopes)
-        elif in_spectra.ndim == 1:
+    # Remove highest output edge, necessary for guvectorize
+    out_edges = out_edges[..., :-1]
+
+    # by default use numpy backend, if requested switch to xarray
+    if xarray is False:
```
jccurtis, line 359:
```suggestion
    if not xarray:
```

jccurtis:
@chunhochow are you comfortable with the extent of the testing here? It would be nice for some limited xarray testing now with an issue for more extensive testing eventually ...

chunhochow:
@jccurtis I'm ok with the entire xarray testing thing being an issue, because it's so isolated from the rest of the code, and I really just need to get this in first (while I don't have much time on this right now).

cosama:
I probably could extend the rebin tests to incorporate `xarray` as well. Would mean we have to add xarray to `requirements-dev.txt` but that would be fine with me.

jccurtis:
@cosama sounds good to me. go for it!

cosama:
@markbandstra @jccurtis Anybody wants to review this? All tests passed and I think it is good to go.

becquerel/core/rebin.py
```diff
@@ -394,63 +301,65 @@ def rebin(in_spectra, in_edges, out_edges, method="interpolation",
         The rebinned spectrum/a
     """
     method = method.lower()
+    if not hasattr(in_spectra, "shape"):
+        assert xarray is False, "Use xarray as input not list"
```
jccurtis, line 305: Maybe I missed this ... where is `xarray` defined? Is this left over?

becquerel/core/rebin.py
```diff
@@ -394,63 +301,65 @@ def rebin(in_spectra, in_edges, out_edges, method="interpolation",
...
-    _check_ndim(in_spectra, {1, 2}, 'in_spectra')
-    _check_ndim(in_edges, {1, 2}, 'in_edges')
-    _check_ndim(slopes, {1, 2}, 'slopes')
-    _check_ndim(out_edges, 1, 'out_edges')
-    # Check shape
-    _check_shape(in_spectra, in_edges, 'in_spectra', 'in_edges', edges=True)
-    _check_shape(in_spectra, slopes, 'in_spectra', 'slopes')
+            # np.rint is a ufunc: natively supported by xarray:
+            in_spectra_rint = np.rint(in_spectra)
+            # TEMP disable this test, always raise warning for now because
```
jccurtis, line 318: We can remove this now right?

tests/rebin_test.py
```diff
@@ -8,6 +8,7 @@
 # ----------------------------------------------
 #         Test Rebinning
 # ----------------------------------------------
+# TODO this only tests np.array inputs now but not xr.DataArray as inputs
```
jccurtis, line 11: This will require `xarray` in `requirements-dev.txt`

cosama: Removed it. Missed that one

becquerel/core/rebin.py
cosama: Yeah left over. Thanks for looking through this so thoroughly. Removed it now.

becquerel/core/rebin.py
cosama: Not sure what part of the code you refer too. I think I would leave the check for `int` type and warn if not. It would be great if we could leave the `isclose` part as well, but as @chunhochow mention in the comment that would break compatibility with data types that only accept `ufuncs`. Thoughts?

jccurtis: I'm referring to the `TEMP` tag which can be resolved before this is merged. I this the `int` type warning is 👍

tests/rebin_test.py
jccurtis: Can you remove the TODO comment as well?

becquerel/core/rebin.py
```diff
@@ -291,73 +261,10 @@ def _rebin_listmode(in_spectrum, in_edges, out_edges, slopes):
...
-    # Init output
-    out_spectra = np.zeros((in_spectra.shape[0], out_edges.shape[0] - 1),
-                           np.int64)
-    for i in np.arange(in_spectra.shape[0]):
-        out_spectra[i, :] = _rebin_listmode(in_spectra[i, :], in_edges[i, :],
-                                            out_edges, slopes[i, :])
-    return out_spectra
+    # N.B. use [:] to assign values to full array output with guvectorize:
+    out_spectrum[:] = np.histogram(energies, bins=out_edges)[0]
+    # handling of overflows no longer needed since the first/last out_edges
```
jccurtis, line 266: Move this comment to the docstring

becquerel/core/rebin.py
```diff
@@ -255,29 +216,38 @@ def _rebin_interpolation(in_spectrum, in_edges, out_edges, slopes):
...
 
     Args:
       in_spectrum: iterable of input spectrum counts in each bin
       in_edges: iterable of input bin edges (len = len(in_spectrum) + 1)
-      out_edges: iterable of output bin edges
-      slopes: unused, just for keeping number of arguments the same as
-              _rebin_interpolation()
+      out_edges_no_rightmost: iterable of output bin edges
+                              (sans the rightmost bin)
+      out_spectrum: for nb.guvectorize; do not actually give this as an arg
```
jccurtis, line 242: Can you clarify:

> do not actually give this as an arg

jccurtis: Also, add a note that this is the same length as `out_edges_no_rightmost`

becquerel/core/rebin.py
```diff
@@ -366,12 +273,12 @@ def rebin(in_spectra, in_edges, out_edges, method="interpolation",
     Spectra rebinning via deterministic or stochastic methods.
 
     Args:
-        in_spectrum (np.ndarray): an array of input spectrum counts (1D or 2D)
-            [num_spectra, num_bins_in] or [num_bins_in]
-        in_edges (np.ndarray): an array of the input bin edges (1D or 2D)
-            [num_spectra, num_bins_in + 1] or [num_bins_in + 1]
+        in_spectrum (np.ndarray): an array of input spectrum counts
```
jccurtis, line 276:
```suggestion
        in_spectrum (np.ndarray): an ND array of input count spectra
```

becquerel/core/rebin.py
```diff
@@ -366,12 +273,12 @@ def rebin(in_spectra, in_edges, out_edges, method="interpolation",
...
+            [..., num_bins_in]
+        in_edges (np.ndarray): an array of the input bin edges
+            [..., num_bins_in + 1] or [num_bins_in + 1]
```
jccurtis, line 279: Add more detail for the two cases (i.e. all the same edges, or different edges matching the number of spectra in)

becquerel/core/rebin.py
```diff
@@ -394,63 +301,64 @@ def rebin(in_spectra, in_edges, out_edges, method="interpolation",
         The rebinned spectrum/a
     """
     method = method.lower()
+    if not hasattr(in_spectra, "shape"):
```
jccurtis, line 304: is this for `xarray`? Otherwise I think the check can be removed.

becquerel/core/rebin.py
jccurtis: Expanding on that: lets either include the below code or remove it altogether. I support a try except if the allclose method doesn't work and a warning that the check cannot be performed. What edge cases would this note work for?

becquerel/core/rebin.py
cosama: Already kind of there. Remove it from here.

becquerel/core/rebin.py
cosama: Done

becquerel/core/rebin.py
markbandstra: i.e., Explain the two cases more explicitly so that someone can more immediately understand what they mean.

cosama: Just fixed this hope it is good now.

cosama:
@jccurtis Please review, once #223 is merged we can update and all tests should pass.

markbandstra (approved):
Great work, let's merge it!

PR #220, opened 2020-12-30
Feature GitHub actions

jccurtis:
Closes #216

- [x] Add basic GH actions test workflow
    - [ ] Make required for merge
- [ ] Remove travis CI 
    - [x] Remove config
    - [ ] Remove travis trigger (on https://travis-ci.org/)
    - [ ] Remove required for merge
- [ ] Format with `black`

PR #221, opened 2020-12-30
Reformat with black

jccurtis:
This should go in before #220 since the CI will not work without the format updates


jccurtis:
Closing in favor of #242

PR #223, opened 2021-01-06
IsotopeQuantity retain floating point precision

### LINKED ISSUE #222: IsotopeQuantity test fails with datetime assert
All recent test have been failing. The error message is:

```
E       AssertionError: assert datetime.datetime(2021, 1, 4, 23, 42, 22, 196186) == datetime.datetime(2021, 1, 4, 23, 42, 34, 811280)
E        +  where datetime.datetime(2021, 1, 4, 23, 42, 22, 196186) = <bound method IsotopeQuantity.time_when of <becquerel.tools.isotope_qty.IsotopeQuantity object at 0x7efd1c4023d0>>(**{'uci': 10.047000000000002})
E        +    where <bound method IsotopeQuantity.time_when of <becquerel.tools.isotope_qty.IsotopeQuantity object at 0x7efd1c4023d0>> = <becquerel.tools.isotope_qty.IsotopeQuantity object at 0x7efd1c4023d0>.time_when
E        +  and   datetime.datetime(2021, 1, 4, 23, 42, 34, 811280) = <becquerel.tools.isotope_qty.IsotopeQuantity object at 0x7efd1c4023d0>.ref_date

```

Anybody has an idea what goes on here. It looks like the python2 test still work, so maybe a version issue with one of the libraries?

cosama:
Closes #222

These changes assure that whatever quantity is used to initalize a `IsotopeQuantity` that quantity will not be affected by floating point rounding errors.

### DIFF
--- a/becquerel/tools/isotope_qty.py
+++ b/becquerel/tools/isotope_qty.py
@@ -6,6 +6,7 @@ import copy
 import numpy as np
 from .isotope import Isotope
 from ..core import utils
+from collections import OrderedDict
 
 UCI_TO_BQ = 3.7e4
 N_AV = 6.022141e23
@@ -98,7 +99,7 @@ class IsotopeQuantity(object):
 
         self._init_isotope(isotope, stability)
         self._init_date(date)
-        self.ref_atoms = self._atoms_from_kwargs(**kwargs)
+        self._ref_quantities = self._quantities_from_kwargs(**kwargs)
 
     def _init_isotope(self, isotope, stability):
         """Initialize the isotope.
@@ -132,8 +133,9 @@ class IsotopeQuantity(object):
             # assume a long-lived source in the current epoch
             self.ref_date = datetime.datetime.now()
 
-    def _atoms_from_kwargs(self, **kwargs):
-        """Parse kwargs and return a quantity in atoms.
+    def _quantities_from_kwargs(self, **kwargs):
+        """Parse kwargs and return a quantity as a OrderedDictionary. The first
+        element in the dictionary is the provided quantity.
 
         Args (specify one):
           atoms: the number of atoms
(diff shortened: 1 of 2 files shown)


becquerel/tools/isotope_qty.py
```diff
@@ -147,24 +149,35 @@ def _atoms_from_kwargs(self, **kwargs):
...
+            bq=lambda : ref_quantities["atoms"] * self.decay_const,
+            uci=lambda : ref_quantities["bq"] / UCI_TO_BQ,
+            g=lambda : ref_quantities["uci"] * UCI_TO_BQ /
+                       self.decay_const / N_AV * self.isotope.A
+        )
+
+        # rotates the order of the list so that the provided kwarg is at [0]
+        order = ["atoms", "bq", "uci", "g"]
+        if next(iter(kwargs)) not in order:
+            raise IsotopeQuantityError("Unknow isotope quantity.")
```
markbandstra, line 173: Unknown

tests/isotope_qty_test.py
```diff
@@ -107,7 +107,10 @@ def test_isotopequantity_ref_atoms_rad(radioisotope, iq_kwargs):
         assert iq.ref_atoms == iq_kwargs['bq'] / radioisotope.decay_const
     else:
         assert iq.ref_atoms == (
-            iq_kwargs['uci'] * UCI_TO_BQ / radioisotope.decay_const)
+            iq_kwargs['uci'] * UCI_TO_BQ / radioisotope.decay_const /
+            N_AV * radioisotope.A / radioisotope.A * N_AV)
+            # Note the last part cancels, but is necessary to get equality with
+            # floating points
```
markbandstra, line 113: This is so weird. Can you explain why this happens, and whether there is a way to avoid it?

becquerel/tools/isotope_qty.py
```diff
@@ -467,18 +505,22 @@ def _mul_div(self, other, div=False):
...
+            **{"date": self.ref_date, key: self._ref_quantities[key] * factor}
+        )
 
     def __eq__(self, other):
         """Equality operation"""
 
         if not isinstance(other, IsotopeQuantity):
             return False
         else:
+            # TODO: Do we need to compare all quantities?
```
markbandstra, line 520: I think probably we test one, and `atoms` seems like a good one, but I am open.

tests/isotope_qty_test.py
cosama: It's one of these weird floating point issues again. The way I do thing is that I calculate `bq` from `atoms`, `uci` from `bq`, `g` from `uci` and then `atoms` from `g`. We always start at the quantity that was provided. In this case `uci` was provided, which is used to calculate `g` which then is used to calculate `atoms`. We have to follow this "path" of derivation to get precisely the same value, which requires the weird conversion to `g` and back.

cosama: It could be avoided by testing with `isclose`, but I would have to check what tolerances would be necessary.

cosama: To be honest I think it would make more sense to test if `iq.quantity_at(quantity, date) == iq_kwargs[quantity]` here (which I think was not always true before and that is why we had the issue), but that seems sort of obvious so I left it.

cosama: And we want to test that the conversion from one unit to the other are right.

becquerel/tools/isotope_qty.py
cosama: Yeah I agree. Just checked, the defaults in `numpy.isclose` are `rtol=1e-05, atol=1e-08`. The absolute tolerance should never be hit here as we handle order of 10^20 atoms, but the relative tolerance is kind of small, we might want to bump this to maybe 1e-08 or so?

becquerel/tools/isotope_qty.py
cosama: Done

tests/isotope_qty_test.py
cosama: I added an `isclose` function, `numpy.isclose` or `math.isclose` both don't work with `uncertanties` as there is a test for `infinity` not supported for that module.

cosama:
I think this would be ready to be pulled in now. Just added the `isclose` fix we discussed earlier and another test for `quantity_at`.

cosama:
I made another change to fix `__eq__` to also work if the isotope quantity is a `uncertanties` object. @markbandstra I assigned you to approve.

markbandstra (approved):

cosama:
@markbandstra Thanks. Will merge once tests pass.

PR #224, opened 2021-01-09
Auto label plot axes in fitting module #218

### LINKED ISSUE #218: Automatic x- and y-axis labels in fitting modules
I typically default to 'counts per bin' and update it later with something like
```py
ax.set_ylabel('counts per {0:.1f} keV'.format(binwidths[0]))
```
I guess here we can get the units from `Spectrum`, too. A new issue for this makes sense to me.

_Originally posted by @jvavrek in https://github.com/lbl-anp/becquerel/pull/146#discussion_r548264499_

jvavrek:
Closes #218


jvavrek:
^ More NNDC errors

jvavrek:
Same notebook plot from https://github.com/lbl-anp/becquerel/pull/146 (very slightly different results here due to bin centers vs edges)

image

jvavrek:
Question: do we want to display the implicit 'per bin' part of the y-axis units?
 - "counts" are really "counts per x bins" or "counts per y keV"
 - "cpskev", if I understand it correctly, is specifically "counts per 1 keV"

becquerel/core/fitting.py
```diff
@@ -338,12 +340,45 @@ def __str__(self):
...
+            '    xmode: {}\n'.format(self.xmode) +
+            '    ymode: {}\n'.format(self.ymode) +
             '       dx: {}\n'.format(self.dx) +
             '      roi: {}'.format(self.roi)
         )
 
     __repr__ = __str__
 
+    @classmethod
+    def from_spectrum(cls, model, spec, xmode, ymode, roi=None):
```
markbandstra, line 352: From meeting: also connect fitting to `Spectrum` with `Spectrum.fit`

jvavrek: Done!

jvavrek:
@cosama in the `SpectrumPlotter.__init__()`:
```
#TODO Marco: maybe we should use tuplets for all the x/y stuff in order
#to reduce the number of attributes (pylint ask for 7 we have 23)!
```
Is this still true? Looks like they have been consolidated into `xlim`, `ylim`, etc.

becquerel/core/fitting.py
cosama, line 352: I am sorry this wasn't clear earlier. I would not have this code here at all, and move it all to `Spectrum.fit`. While this works sort of, it is a easy source of circular dependencies (for example if we want to add a test that this really only can take `Spectrum` objects we would need to `import spectrum`. I don't see a use-case where this would be useful and `Spectrum.fit` can't accomplish the same thing.

cosama:
> @cosama in the `SpectrumPlotter.__init__()`:
> 
> ```
> #TODO Marco: maybe we should use tuplets for all the x/y stuff in order
> #to reduce the number of attributes (pylint ask for 7 we have 23)!
> ```
> 
> Is this still true? Looks like they have been consolidated into `xlim`, `ylim`, etc.

I think this is still true (I haven't been running pylint for a while). I think it includes `self._*` in the count. This has never been strictly enforced though and probably won't even be checked once we move to `black`. We can probably just remove the comment there. It is not optimal but I don't see a point in changing it just because some weird spec doesn't like it :-)

becquerel/core/fitting.py
jvavrek: Makes sense. I'll fix that now.

jvavrek: Resolving based on the latest commit

markbandstra (approved):
Approved, good work

PR #230, opened 2021-02-12
Add a simple plot() method to Fitter

jvavrek:
The `Fitter` class currently lacks a simple `plot` method for displaying just the fit result, especially if one wants to quickly overlay it on an existing plot. This provides such a method.


markbandstra (approved):
Nice quick PR

Issue #180, opened 2019-07-18
Store energy calibration in Spectrum object


jccurtis:
As mentioned in #179 we needed to resolve this oversight in the calibration routine.

jvavrek:
This was closed by #231

PR #231, opened 2021-02-17
Store EnergyCal in Spectrum

jvavrek:
Fix the auto merge of #229 

> I'm opening this since I think we talked about getting #180 soon as well. Is there anything other than simply adding a self.cal to Spectrum? Perhaps the rebin or combine_bins change the calibration, depending on how the calibration is defined?
> 
> Thoughts @jccurtis @markbandstra @cosama ?

### DIFF
--- a/becquerel/core/spectrum.py
+++ b/becquerel/core/spectrum.py
@@ -140,7 +140,7 @@ class Spectrum(object):
         self._cps = None
         self._bin_edges_kev = None
         self._bin_edges_raw = None
-        self.cal = None
+        self.energy_cal = None
         self.livetime = None
         self.realtime = None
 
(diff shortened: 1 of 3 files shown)


becquerel/core/spectrum.py
```diff
@@ -140,6 +140,7 @@ def __init__(self, counts=None, cps=None, uncs=None, bin_edges_kev=None,
         self._cps = None
         self._bin_edges_kev = None
         self._bin_edges_raw = None
+        self.cal = None
```
markbandstra, line 143: Rename to `energy_cal` since we will add a resolution calibration later.

jvavrek:
Probably should add a test or two

jvavrek:
@markbandstra I think this should be good to go once the pipeline passes

markbandstra (approved):
Looks good!

PR #235, opened 2021-02-18
Let date be a date, not just datetime; also default _at() methods to now()

### LINKED ISSUE #226: IsotopeQuantity has a kwarg `date` that cannot be a `datetime.date` object
The kwarg `date` in IsotopeQuantity (and its docstring) suggest that a `datetime.date` object should work, but it won't:
(note for posterity that the actual calibration date of this source is Apr 1, 2019)
```
>>> bq.IsotopeQuantity('Cs-137', uci=189.6, date=datetime.date(2020, 11, 5))

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-90-1eb84c4f3676> in <module>
----> 1 bq.IsotopeQuantity('Cs-137', uci=189.6, date=datetime.date(2020, 11, 5))

~/becquerel/becquerel/tools/isotope_qty.py in __init__(self, isotope, date, stability, **kwargs)
     99 
    100         self._init_isotope(isotope, stability)
--> 101         self._init_date(date)
    102         self._ref_quantities = self._quantities_from_kwargs(**kwargs)
    103 

~/becquerel/becquerel/tools/isotope_qty.py in _init_date(self, date)
    129 
    130         self.ref_date = utils.handle_datetime(
--> 131             date, error_name='IsotopeQuantity date', allow_none=True)
    132         if self.ref_date is None:
    133             # assume a long-lived source in the current epoch

~/becquerel/becquerel/core/utils.py in handle_datetime(input_time, error_name, allow_none)
    130     else:
    131         raise TypeError(
--> 132             'Unknown type for {}: {}'.format(error_name, input_time))
    133 
    134 

TypeError: Unknown type for IsotopeQuantity date: 2020-11-05
```

Instead it requires a`datetime.datetime` object, i.e., an additional time specification. Can we just set the time to 12 noon like I think most of these sources do, and maybe throw a warning when doing so?

jvavrek:
Closes #226

### DIFF
--- a/tests/isotope_qty_test.py
+++ b/tests/isotope_qty_test.py
@@ -265,6 +265,11 @@ def test_isotopequantity_activity_now(iq):
     with pytest.warns(DeprecationWarning):
         assert np.isclose(iq.g_now(), iq.g_at(datetime.datetime.now()))
 
+    assert np.isclose(iq.bq_at(), iq.bq_at(datetime.datetime.now()))
+    assert np.isclose(iq.uci_at(), iq.uci_at(datetime.datetime.now()))
+    assert np.isclose(iq.atoms_at(), iq.atoms_at(datetime.datetime.now()))
+    assert np.isclose(iq.g_at(), iq.g_at(datetime.datetime.now()))
+
 
 def test_isotopequantity_decays_from(iq):
     """Test IsotopeQuantity.*_from()"""
--- a/tests/isotope_qty_test.py
+++ b/tests/isotope_qty_test.py
@@ -256,10 +256,14 @@ def test_isotopequantity_time_when_error(stable_isotope):
 def test_isotopequantity_activity_now(iq):
     """Test IsotopeQuantity.*_now()"""
 
-    assert np.isclose(iq.bq_now(), iq.bq_at(datetime.datetime.now()))
-    assert np.isclose(iq.uci_now(), iq.uci_at(datetime.datetime.now()))
-    assert np.isclose(iq.atoms_now(), iq.atoms_at(datetime.datetime.now()))
-    assert np.isclose(iq.g_now(), iq.g_at(datetime.datetime.now()))
+    with pytest.warns(DeprecationWarning):
+        assert np.isclose(iq.bq_now(), iq.bq_at(datetime.datetime.now()))
+    with pytest.warns(DeprecationWarning):
+        assert np.isclose(iq.uci_now(), iq.uci_at(datetime.datetime.now()))
+    with pytest.warns(DeprecationWarning):
+        assert np.isclose(iq.atoms_now(), iq.atoms_at(datetime.datetime.now()))
+    with pytest.warns(DeprecationWarning):
+        assert np.isclose(iq.g_now(), iq.g_at(datetime.datetime.now()))
 
 
 def test_isotopequantity_decays_from(iq):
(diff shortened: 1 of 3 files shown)


cosama:
Looks good to me. Might make sense to add a test for the `datetime.date` thing.

Also there is this test: https://github.com/lbl-anp/becquerel/blob/update-date/tests/isotope_qty_test.py#L249. We could add no argument to it.

Looking at that I realized that there is a `*_now` method as well. Maybe can deprecate that and simplify the interface by defaulting to now with the `*_at` methods. @markbandstra Any thoughts?

jvavrek:
@cosama this should be good to go:
 - added a test for the `datetime` parse
 - added no argument to `test_isotopequantity_activity_now()`
 - deprecated the `_now()` methods

tests/isotope_qty_test.py
```diff
@@ -244,10 +256,19 @@ def test_isotopequantity_time_when_error(stable_isotope):
...
+    with pytest.warns(DeprecationWarning):
+        assert np.isclose(iq.bq_now(), iq.bq_at(datetime.datetime.now()))
+    with pytest.warns(DeprecationWarning):
+        assert np.isclose(iq.uci_now(), iq.uci_at(datetime.datetime.now()))
+    with pytest.warns(DeprecationWarning):
+        assert np.isclose(iq.atoms_now(), iq.atoms_at(datetime.datetime.now()))
+    with pytest.warns(DeprecationWarning):
+        assert np.isclose(iq.g_now(), iq.g_at(datetime.datetime.now()))
+
+    assert np.isclose(iq.bq_at(), iq.bq_at(datetime.datetime.now()))
```
cosama, line 268: These tests might break on very slow machine. They only work right now because `rtol=1e08` by default for isclose. I think it would be better to define an absolute tolerance as a function of of how much decay we would expect in lets say a second.

```
assert np.isclose(iq.bq_at() - iq.bq_at(datetime.datetime.now()), iq.bq_during(iq.ref_time, iq.ref_time + datetime.timedelta(0,1)))
```
Not very pretty but probably safer. Feel free to change this if you see a better way of implementing it.

jvavrek: Ah yeah, I figured this could be the case. I'll think about this this afternoon.

jvavrek: The issue with using 1 s is that 1 s can mean very different activities at `now` vs at the `ref_time`

jvavrek: We could possibly do a speed test and `@pytest.mark.skipif` if the two `datetime.datetime.now()` are too different. Or we could just loosen the `rtol` in `np.isclose`

cosama: `rtol` is very generous for large numbers, not sure if it is needed. Maybe it is fine as it is.

cosama:
just realize there is only `ref_atoms` but no `ref_bq` or other units. Probably worth adding that? Here or new one?

tests/isotope_qty_test.py
markbandstra: How about this sequence:
```
x0 = iq.bq_at()
x1 = iq.bq_at(datetime.datetime.now())
x2 = iq.bq_at()
```
and test that `x2 <= x1 <= x0`.

jvavrek: done

markbandstra (approved):
Thanks!

jvavrek:
> just realize there is only `ref_atoms` but no `ref_bq` or other units. Probably worth adding that? Here or new one? Maybe a `ref_quantity` as well?

The `ref_atoms` docstring says:
```
Access the reference atoms directly (for backwards compatibility)
```
so I assume that `self._ref_quantities` is the preferred way.

PR #237, opened 2021-02-23
add specific_activity prop to bq.Isotope

jvavrek:
Useful for converting mass of material to activity


cosama (approved):
Looks good to me.

Issue #219, opened 2020-12-23
Improve initial guesses in fitting module?


jvavrek:
@jccurtis those are with the old guess function. We can look into improving guesses in the future.

_Originally posted by @jvavrek in https://github.com/lbl-anp/becquerel/issues/146#issuecomment-750483392_

For instance, using the max bin can give a closer height estimate than the sum of all the counts in the ROI. See my comment in the aforementioned PR: https://github.com/lbl-anp/becquerel/pull/146#issuecomment-749776169

However, this is not too crucial, as long as the fits still converge well.

jvavrek:
This was closed by #238

PR #238, opened 2021-02-26
improve peak amplitude guess for gauss, gausserf, expgauss

### LINKED ISSUE #225: Initial guess of fitting can be improved for Gaussian
The intial guess for fitting seems to be off by about a factor of 2 or so for the gaussian case. I never managed to track this down, but I assume it is related to some normalization with the bin with or a factor of `2*pi`.

jvavrek:
Closes #225 and #219


markbandstra (approved):
Looks good!

PR #239, opened 2021-02-27
plot the fit evaluation error bands, and ensure fit is shown above other lines

jvavrek:
Quick PR to add error bounds from the fit to `custom_plot`, and also ensure that the fit line and error band are drawn above other lines

image


jvavrek:
Note that the sigma residuals still just use the fit eval, and do not factor in the fit uncertainty—only the data uncertainty.

markbandstra (approved):
Rubber stamp!

PR #248, opened 2021-03-16
Refactor and simplify calibrations

### LINKED ISSUE #245: Refactoring and simplifying calibrations
I propose refactoring and simplifying our calibrations by creating a new, simpler base class for all calibrations of any kind. At their essence, calibrations are simply functions that map some scalar quantity (raw channels or energy in keV) to some scalar quantity. Specific types of calibrations are energy calibrations (raw->keV), energy resolution calibrations (keV->keV), and efficiency calibrations (keV->efficiency).

The base class should handle:
* Instantiation with some set of scalar parameters that uniquely define the scalar function
* HDF5 I/O with read and write methods (see #244)

Any subclass needs to:
* Evaluate some scalar function at a scalar argument, perhaps as a `__call__` method so the instance itself can be called like a function
* Implement a `from_points` method to fit the function to a set of calibration points, which are optional and stored for future reference. Might also want to consider fit weights as an optional argument (#133).

For example, we could have a `LinearCalibration` class that just evaluates a linear function for a set of coefficients, or a `PolynomialCalibration` that does the same for a higher order polynomial. The classes can be agnostic to whether they are being used as an energy calibration or some other kind of calibration.

If you look at `LinearEnergyCal`, there seems to be a lot of bloat surrounding the naming and accessing of coefficients. Why not just have them as arrays of floats defined in the docstrings? I think it would make it easier to develop and implement new kinds of calibrations with less overhead.

This change would address #32.

I vote to close #45, that seems too custom of an application.

markbandstra:
Adds a class for generic calibrations called `Calibration`, in addition to subclasses for specific types of calibration functions.

To create an arbitrary calibration function, simply write the function as a string, where `x` is the independent variable and `p` are the parameters, each of which must be referenced explicitly with an integer index. For example, here is how to instantiate a linear calibration with intercept of 5 and slope of 2:

```python
cal = Calibration("p[0] + p[1] * x", [5, 2])
```

Here is a linear calibration with no intercept and a slope of 2:

```python
cal = Calibration("p[0] * x", [2])
```

Calibrations are callable like functions, so `cal(x)` will return its value at `x`:

```python
cal(10)
# Out: 20.0
```

Calibrations can include special functions, basically anything that [`asteval`](https://newville.github.io/asteval/) can handle, including many `numpy` functions:

```python
cal = Calibration('p[0] + p[1] * x + p[2] * np.sqrt(x) + p[3] * np.exp(x / p[4])', [5, 2, 3, 7, 1000], domain=[0, 3000])
```

To fit a calibration function to points, you can use the `set_points` and `add_points` methods and then call `fit`, or instantiate the class using `from_points`:

```python
cal = Calibration.from_points("p[0] * x", [0, 1000], [0, 200])
cal.params
# Out: array([0.2])
```

There are initialization methods for `Calibration` that are designed to make the creation of different functional forms easier. For example, here is a 4th order polynomial, where the order is inferred from the number of parameters:

```python
cal = Calibration.from_polynomial([5, 2, 4, 1, 3])
cal.expression
# Out: 'p[0] + p[1] * x ** 1 + p[2] * x ** 2 + p[3] * x ** 3 + p[4] * x ** 4'
```

One of the initialization methods generates an expression that interpolates between the calibration points:

```python
cal = Calibration.from_interpolation([0, 1000], [0, 200])
print(cal.expression)
# scipy.interpolate.interp1d([0, 1000], [0, 200], fill_value="extrapolate")(x)
```

This PR also adds tools to perform HDF5 I/O, which is fully supported for `Calibration` instances. So for any calibration you can do
```python
cal.write("filename.h5")
```
and it will just work (`Calibration.read("filename.h5")` as well). You can also give an argument that is an already open HDF5 file or group.

Closes #245 
Commits were cherry-picked from #244

TODO:
- [x] add basic tools for HDF5 I/O
- [x] add new `Calibration` class
- [x] implement `fit` and `from_points`
- [x] implement `__str__` and `__repr__`
- [x] pass in bounds and any other fitting parameters to `fit()` and `from_points()`
- [x] allow the user to specify weights for each calibration point for use by `fit()`
- [x] decide how strict we are going to be about extrapolation beyond calibration points
- [x] determine how to organize `Calibration` and its subclasses within `becquerel` --- import them all to the top level?
- [x] replace all occurrences of `LinearEnergyCal` with the new `LinearCalibration`
- [x] update the `energycal` example notebook


becquerel/core/calibration.py
```diff
@@ -0,0 +1,787 @@
...
+import copy
+import asteval
+import black
+import blib2to3
+import numpy as np
+import scipy
+import scipy.optimize
+from .. import io
+
+CLIP_MAX = 1e6  # maximum value for a calibration function
```
jvavrek, line 14: This could be improved by implementing domains, I think.

becquerel/core/calibration.py
```diff
@@ -0,0 +1,787 @@
...
+        raise CalibrationError(f"All calibration x points must be >= 0: {points_x}")
+    if not np.all(points_y >= 0):
+        raise CalibrationError(f"All calibration y points must be >= 0: {points_y}")
+    return points_x, points_y
+
+
+class Calibration(object):
+    """Base class for calibrations.
+
+    A calibration is a nonnegative scalar function of a nonnegative scalar
```
jvavrek, line 78: This is not quite true if e.g. a linear calibration has a negative offset.

markbandstra: Yes, I noticed this later when coding -- I had to remove a non-negativity check on the function output.

jvavrek:
One other request: fit statistics. R^2 and (reduced) chi2 are the big ones. I can't remember if `scipy.optimize.least_squares()` provides those or we would need to do it ourselves.

becquerel/core/calibration.py
```diff
@@ -0,0 +1,787 @@
...
+            raise CalibrationError(
+                f"asteval failed with errors:\n"
+                + "\n".join(str(err.get_error()) for err in safe_eval.error)
+            )
+        if not np.all(np.isreal(y)):
+            raise CalibrationError(
+                f"Function evaluation resulted in complex values: {y}"
+            )
+        # clip values of y
+        y = np.clip(y, 0, CLIP_MAX)
```
jvavrek, line 171: @markbandstra we should remove this as well. I have a linear calibration with a negative intercept, so the first few bin edges in keV are negative, and all get clipped to 0. Then applying the calibration causes `ValueError: Bin edge energies must be strictly increasing`.

markbandstra: Probably means we need both an x- and y-domain specified then.

markbandstra:
TODO items from a recent meeting:

- [x] move static methods to hidden module-level functions
- [x] make subclasses factory functions
- [x] fix logic in `from_points` -- `include_origin` does not work right
- [x] use `expression` everywhere
- [ ] specify and enforce function domain and range
- [ ] provide access to optional/auxiliary parameters ("q"?)
- [ ] specify whether expected to be invertible
- [ ] if invertible, provide inverse expression
    - calculate it automatically if none, but will be slower
    - use scipy root finding
- [ ] make `Calibration` into `EquationCalibration`, and have an abstract base class
- [ ] allow named parameters
    - parameters are a dict instead of an array

becquerel/core/calibration.py
markbandstra: Should be resolved now with the addition of `domain` and `range` to `Calibration`.

becquerel/core/calibration.py
markbandstra: Should be resolved now with the addition of `domain` and `range` to `Calibration`.

becquerel/core/calibration.py
markbandstra: Just fixed this docstring.

jvavrek:
I don't know if this is necessarily a responsibility for `becquerel`, but a useful feature would be the option to scale the x/y inputs (x especially) before performing the calibration, and then rescaling the calibration back after the fact. For various reasons I'm trying out a high-order polynomial fit and running into overflows when the input ADC value is > 4000 or so. I can divide by 1000 on my end, but keeping track of whether different calibrations use the downscaled values will be a pain.

markbandstra:
> For various reasons I'm trying out a high-order polynomial fit and running into overflows when the input ADC value is > 4000 or so.

Does this happen when you are calling `fit`? If so you can send the `bounds` kwarg to limit the range of the offending parameter(s).

jvavrek:
> > For various reasons I'm trying out a high-order polynomial fit and running into overflows when the input ADC value is > 4000 or so.
> 
> Does this happen when you are calling `fit`? If so you can send the `bounds` kwarg to limit the range of the offending parameter(s).

It's when I call `Calibration.from_points()`. I'm not sure where exactly inside there that things go awry. I don't think the bounds are the issue, I think it is just the high-order polynomial and ADC input values in the thousands. E.g., I can get away with
```
p[0] + p[1] * x ** 1 + p[2] * x ** 2 + p[3] * x ** 3
```
but for the next higher order I need to do something like
```
p[0] + p[1] * (x/1000) ** 1 + p[2] * (x/1000) ** 2 + p[3] * (x/1000) ** 3 + p[4] * (x/1000) ** 4
```

markbandstra:
I would try setting the bounds, and if that still doesn't solve it you can always put the scaling directly into the expression, as you have above:

```python
p[0] + p[1] * (x/1000) ** 1 + p[2] * (x/1000) ** 2 + p[3] * (x/1000) ** 3 + p[4] * (x/1000) ** 4
```

is a valid expression. Also check that the new property `domain` is limited to (0, 4000) -- the default domain is much wider and it will test values throughout the entire domain when you instantiate the calibration.

becquerel/core/calibration.py
```diff
@@ -637,15 +748,18 @@ def __eq__(self, other):
...
 
     def copy(self):
         """Make a complete copy of the calibration."""
         cal = Calibration(
             self.expression,
             self.params,
+            aux_params=self.aux_params,
             inv_expression=self.inv_expression,
             domain=self.domain,
             rng=self.range,
```
jvavrek: Do any of these args need `.copy()`?

becquerel/core/spectrum.py
```diff
@@ -1070,16 +1070,13 @@ def get_bin_properties(self, use_kev=None):
...
 
         Args:
-          cal: an object derived from EnergyCalBase
+          cal: a Calibration object
         """
 
-        try:
-            self.bin_edges_kev = cal.ch2kev(self.bin_edges_raw)
-        except AttributeError:
-            self.bin_edges_kev = cal(self.bin_edges_raw)
```
jvavrek: Should we maintain backwards-compatibility for a while?

jvavrek (changes requested):
Looks pretty good, just those two above comments

becquerel/core/calibration.py
markbandstra: I don't think so -- `params` and `aux_params` are copied during `__init__` (it's hidden in the setters like [this one](https://github.com/lbl-anp/becquerel/blob/new-calibration/becquerel/core/calibration.py#L635)). For `attrs` I perform a deep copy [here in its setter](https://github.com/lbl-anp/becquerel/blob/new-calibration/becquerel/core/calibration.py#L675). The others are strings or tuples.

becquerel/core/spectrum.py
markbandstra: Yeah, probably, I was too zealous. I'll restore the old ways and throw a `DeprecationWarning` when they are used.

jvavrek (approved):

PR #250, opened 2021-03-19
Add plot method to EnergyCalBase

jvavrek:
This adds some basic plots:

1. scatter plot of input energy vs input channel
2. line plot of fit energy vs input channel
3. scatter plot of (fit energy - input energy) vs channel

Examples from `energycal.ipynb` below.

This is mostly so one can do a quick `cal.plot()` instead of rewriting a bunch of plotting.


jvavrek:
From `energycal.ipynb`:

If the calibration has points:

image

If it does not:

image

jvavrek:
@jccurtis @markbandstra hopefully this does not conflict with plans you had for refactoring calibrations, but if it will, let me know!

markbandstra:
> @jccurtis @markbandstra hopefully this does not conflict with plans you had for refactoring calibrations, but if it will, let me know!

No problem, I think it is a good idea and will work it into #245 as it goes forward.

markbandstra (approved):
Looks good to me

PR #251, opened 2021-03-20
Fitting with iminuit

jvavrek:
Currently we can do least-squares fitting with errors, or Poisson maximum likelihood fitting without errors. I kind of need Poisson maximum likelihood with errors, which is something that iminuit can do via the Minos error estimation routine.

This PR hacks together some lmfit/iminuit compatibility. I.e., it uses our existing lmfit-oriented code from @jccurtis and @cosama, and translates it into something iminuit can use. They have somewhat different interfaces, so there are a couple of potentially screwy hax that might need cleaning up.

Ultimately, we might want to consider yet another layer of abstraction, having a `Fitter` base class and two `LMFitter` and `MinuitFitter` derived classes. What happens to `class FitterGaussGauss(Fitter)` then is an open question.

~There are still some issues to resolve regarding parameter values. Currently, iminuit prints wrong parameter values, but the plots look correct. To be determined!~ Edit: nope, iminuit is correct, and I was forgetting the sqrt(2pi) in the gaussian amplitude


becquerel/core/fitting.py
```diff
@@ -604,13 +618,17 @@ def fit(self, backend="lmfit"):
         y_roi_norm = self.y_roi
         if self.dx is not None:
             y_roi_norm = self.y_roi / self.dx_roi
-        if backend.lower().strip() == "lmfit":
+
+        self.backend = backend.lower().strip()
+
+        if self.backend == "lmfit":
             # Perform the fit, weighted by 1/uncertainties.
             weights = self.y_unc_roi ** -1.0
```
jvavrek, line 626: @cosama aren't the weights normally 1/uncertainties^2 ? Or is this an `lmfit` convention?

becquerel/core/fitting.py
```diff
@@ -624,6 +642,64 @@ def fit(self, backend="lmfit"):
...
+            # Supplying guesses and/or bounds seems more important for minuit
+            # TODO: are some already specified?
+            if limits is not None:
+                for k, v in limits.items():
+                    self.result.limits[k] = v
+
+            # Run the minimizer
+            self.result.migrad()
+
+            # Update the interface to best_values
```
jvavrek, line 693: Maybe this goes after error calc?

becquerel/core/fitting.py
```diff
@@ -663,26 +739,44 @@ def param_val(self, param):
...
-            raise FittingError("Unknown param: {}", param)
+            raise FittingError("Unknown backend: {}", self.backend)
+
+    @property
+    def best_values(self):
+        """Wrapper for dictionary of best_values."""
+        try:
+            return self.result.best_values
+        except AttributeError:
+            return self._best_values
```
jvavrek, line 779: We can now check `self.backend` instead of `try/except`

becquerel/core/fitting.py
cosama: Good question. Would have too think about that. Either way ` 1 / self.y_unc_roi ** -1.0` should be faster.

jvavrek:
Here's the motivating example. Take a pure-Gaussian peak of say 1000 counts:
```py
np.random.seed(2021)
samples = np.random.normal(size=1000)
counts, edges = np.histogram(samples, bins=100)
```

Fit it with `backend=lmfit`, which uses least-squares loss:
```py
fitter = bq.Fitter(
    model=['gauss'],
    x=edges[:-1],
    y=counts,
    y_unc=np.sqrt(counts)
)
fitter.fit(backend='lmfit')
```

The fit looks visually OK, sure:
image

But the fit counts are biased low, by 7.5%, which is almost 3 sigma:
```py
fitter.param_val('gauss_amp') / np.diff(edges)[0]  # 925.97
fitter.param_unc('gauss_amp') / np.diff(edges)[0]  # 27.58
```

If we fit with `backend='lmfit-pml'`, things improve, but we can't get error estimates:
image

```py
fitter.param_val('gauss_amp') / np.diff(edges)[0]  # 1000.42
fitter.param_unc('gauss_amp') / np.diff(edges)[0]  # TypeError: unsupported operand type(s) for /: 'NoneType' and 'float'
```

But now if we fit with `backend='minuit'`:
image

```py
fitter.param_val('gauss_amp') / np.diff(edges)[0]  # 1001.32
fitter.param_unc('gauss_amp') / np.diff(edges)[0]  # 31.67
```

We get a fit uncertainty on the gaussian amplitude, and in this simple case, it is very close to the sqrt of the fit counts.

For a more complicated case, see the updated `fitting.ipynb` notebook, which shows how to specify parameter bounds and limits as well.

jvavrek:
@cosama I'd like to add `"minuit"` to the list of methods in `fitting_test.py`, but I keep getting KeyErrors for `expgauss_fwhm`. I thought I solved this for the regular `gauss_fwhm` but I guess not. Would it be possible to just get rid of the `*_fwhm` parameters? They're a pain to handle and should be easy enough to get from the fit sigma result.

tests/fitting_test.py
```diff
@@ -43,18 +43,14 @@ def sim_data(x_min, x_max, y_func, num_x=200, binning="linear", **params):
...
+        assert np.isclose(v, fit_params[p], rtol=rtol), p
 
 
 def compare_counts(fitter):
     data_counts = np.sum(fitter.y_roi)
     model_counts = np.sum(
-        fitter.eval(fitter.x_roi, **fitter.result.best_values) * fitter.dx_roi
+        fitter.eval(fitter.x_roi, **fitter.best_values) * fitter.dx_roi
     )
     assert np.allclose(data_counts, model_counts, atol=1e-2)
```
jvavrek, line 55: Is this correct? We're comparing both the model counts and data counts with their background contributions. If the background model is bad, the test will fail, despite the peak counts matching.

markbandstra: Not sure if this is helpful or even what you are going for here, but it will always be true for Poisson maximum likelihood that, if the model you are fitting is able to scale itself so that sum(model) = sum(counts), then it will always do so.

cosama: Yeah, this is one of the features of Poisson likelihood. As long as the range is the same the counts in the model and the data will be identical. If that is not the case then the fit is not a proper Poisson fit. I thought it would be a good thing to test to make sure that the fit indeed is Poisson in nature.

cosama (commented):
@jvavrek I have no strong feelings about guass_fwhm. We can remove it. If you post the KeyError that you get I might also be able to figure out what is going wrong there.

jvavrek:
> @jvavrek I have no strong feelings about guass_fwhm. We can remove it. If you post the KeyError that you get I might also be able to figure out what is going wrong there.

It was throwing a KeyError in the tests. I worked around it by iterating over `true_params` instead of `fit_params`:
```py
def compare_params(true_params, fit_params, rtol, fitter):
    for p, v in true_params.items():
        assert np.isclose(v, fit_params[p], rtol=rtol), p
```

tests/fitting_test.py
jvavrek: 😎

becquerel/core/fitting.py
```diff
@@ -624,6 +647,116 @@ def fit(self, backend="lmfit"):
...
+                if self.dx_roi is not None:
+                    y_eval *= self.dx_roi
+                lp = poisson_loss(y_eval, self.y_roi)
+                return lp
+
+            # TODO: we could add the option for iminuit with least-squares
+
+            # Filter out fixed params and have one consistent variable name
+            # instead of all the params / param_names / parameters / etc.
+            free_vars = [p for p in self.params if self.params[p].vary]
```
markbandstra, line 673: `free_vars` and `self.params.keys()` set the ordering of the variables. Make sure other dicts/lists follow the same order.

becquerel/core/fitting.py
```diff
@@ -624,6 +647,116 @@ def fit(self, backend="lmfit"):
...
+                # _requires_ args.
+                # TODO: check if the above is true for Minuit
+                # TODO: this feels dangerous!
+                # TODO: test out-of-order kwargs
+                kwargs = {self.model.param_names[i]: arg for i, arg in enumerate(args)}
+                y_eval = self.model.eval(x=self.x_roi, **kwargs)
+                if self.dx_roi is not None:
+                    y_eval *= self.dx_roi
+                lp = poisson_loss(y_eval, self.y_roi)
+                return lp
```
markbandstra, line 667: Support least squares loss as well?

becquerel/core/fitting.py
```diff
@@ -624,6 +647,116 @@ def fit(self, backend="lmfit"):
...
+                # while (I think) the cost function passed to the Minuit object
+                # _requires_ args.
+                # TODO: check if the above is true for Minuit
+                # TODO: this feels dangerous!
+                # TODO: test out-of-order kwargs
+                kwargs = {self.model.param_names[i]: arg for i, arg in enumerate(args)}
+                y_eval = self.model.eval(x=self.x_roi, **kwargs)
+                if self.dx_roi is not None:
+                    y_eval *= self.dx_roi
+                lp = poisson_loss(y_eval, self.y_roi)
```
markbandstra, line 666: We should warn if the data are not counts (e.g., subtracted spectrum).

becquerel/core/fitting.py
markbandstra: `lmfit` seems to define weights as inverse of standard deviation. From their [documentation](https://lmfit.github.io/lmfit-py/model.html):

```
weights (array_like, optional) – Weights to multiply (data-model) for fit residual.
```

So the weights have to have units of "inverse data"; in other words, least squares will minimize `sum(weights**2 * (data - model)**2`.

Another way to see confirm this is that `ModelResult.plot` will plot errorbars in `y` if `yerr` is given, but if not:

```
If yerr is not specified and the fit includes weights, yerr set to 1/self.weights.
```

becquerel/core/fitting.py
jvavrek: @markbandstra since we don't have access to the full `Spectrum` at this point, I wrote a little test:
```py
def _is_count_like(y):
    """Determine whether data 'looks like' it could be counts.

    Parameters
    ----------
    y : array-like
        Y-data
    """
    if np.any(y) < 0:
        return False
    if not np.allclose(y.astype(np.int), y):
        return False
    return True
```

Then I raise a `FittingWarning` if non-count-like data is passed into the Poisson loss fitter.

markbandstra: Should it be `np.any(y < 0)`?

jvavrek: Yes, thanks

markbandstra: Seems like a sensible way to test for count data, setting aside a larger solution that involves tracking a "count" state within `Spectrum`.

cosama:
```
(y + 0.5).astype(int)
```

I don't think we need np.int specifically. Also this prevents weird rounding errors if the float is just a bit below the actual int.

cosama: It feels like we had this discussion before, but I couldn't find it in a quick search. The fitting module uses:

```
np.rint(in_spectra).astype(int)
```

jvavrek: Ok, I like the `np.rint` solution.

becquerel/core/fitting.py
jvavrek: I suspect that under the hood the fitter would then use weights^2, but that's not an issue for here.

becquerel/core/fitting.py
markbandstra: Just raise `NotImplementedError` for now and note that `lmfit` can handle this case.

jvavrek:
@cosama @markbandstra I think everything is resolved here. Good to go?

cosama:
I still think it would make sense to split up some of the functions, but I really want to start using this, so I think it makes more sense to merge this and maybe I can put some time aside to polish it out some (i.e. bring the different backends into a more consistent structure).

cosama (approved):
I approve for now. Maybe @markbandstra has some other comments.

markbandstra (approved):
Good work!

PR #252, opened 2021-03-23
Constant min 0

jvavrek:
Hotfix: sets a minimum of zero on a ConstantModel. Useful when using Poisson likelihood to avoid nan's


markbandstra (approved):
Looks good! We should think about these conditions for other models as well.

PR #254, opened 2021-03-27
Fix FWHM formula used by `PeakFilter`

markbandstra:
A user with a keen eye noticed that the `PeakFilter` class has been using an erroneous formula for the spectral energy resolution:

```python
f(x) = np.sqrt(f0 ** 2 + (f1 ** 2 - f0 ** 2) * (x / x1) ** 2)
```

which interpolates between `f(0) = f0` and `f(x1) = f1`, but it scales roughly as `x`, not the intended `sqrt(x)`.

This formula has been changed to

```python
f(x) = np.sqrt(f0 ** 2 + (f1 ** 2 - f0 ** 2) * (x / x1))
```

and the notebook has been brought up to date.


jvavrek (approved):
looks good!

markbandstra:
I should also note that this appears to be the only location where this bug crops up. This change may require users to adjust existing code that uses the `autocal` tools, although the `PeakFilter` kernels are pretty robust (which is why we never noticed that problem) so it is also possible that no changes will be needed.

jvavrek:
Ah, autocal tests fail

markbandstra:
Dangit, always run the tests. I'll look into it.

jvavrek (approved):

PR #259, opened 2021-04-06
Peakfinder: fix index error by ignoring peaks too close to the edge

arunpersaud:
Quick fix to just ignore peaks close to the edges (got a few IndexErrors on my data)


becquerel/core/peakfinder.py
```diff
@@ -265,6 +265,11 @@ def add_peak(self, xpeak):
             fwhm0 = self.kernel.fwhm(xpeak)
             bw = self.spectrum.bin_widths_raw[0]
             h = int(max(1, 0.2 * fwhm0 / bw))
+
+            # skip peaks that are too close to the edge
+            if (xbin - h < 0) or (xbin + h > len(self.snr) - 1):
+                return
```
jvavrek, line 271: Can you add a warning that this peak is being skipped?

arunpersaud: Sure. I added a warning via a print statement. Wasn't quite sure what format you use... let me know if this one is OK.

jvavrek: We typically use `warnings.warn()`. There should be an instance of that in the `peakfinder.py` file. In this case we can probably leave the warning type empty (defaults to a `UserWarning` I think) instead of a `DeprecationWarning`.

arunpersaud: changed it and created a peakfindingwarning

becquerel/core/peakfinder.py
```diff
@@ -14,6 +14,12 @@ class PeakFilterError(Exception):
     pass
 
 
+class PeakFindingWarning(UserWarning):
```
markbandstra, line 17: Could you rename this `PeakFinderWarning` to be consistent with `PeakFinderError`? Might also make sense to move it down near `PeakFinderError` around line 155.

markbandstra: @arunpersaud any chance you could make this change? Thanks!

arunpersaud:
Just pushed an update

markbandstra (approved):
Looks great, thanks for fixing this bug!

PR #260, opened 2021-04-06
Precommit

arunpersaud:
In case you are interested in using pre-commit for running black. Perhaps adding pyupgrade would also be useful


markbandstra:
This PR looks good. It looks like we need to wait for #259 before approval and merging.

markbandstra (approved):
Great idea, we didn't know about this tool. Thanks also for updating the CONTRIBUTING instructions.

PR #261, opened 2021-04-06
ran pyupgrade --py36-plus over the codebase

arunpersaud:
Found some future imports that didn't look as if they were needed anymore and in the end thought I just run pyupgrade over it.

### DIFF
--- a/becquerel/core/autocal.py
+++ b/becquerel/core/autocal.py
@@ -222,7 +222,7 @@ def find_best_gain(
         }
 
 
-class AutoCalibrator(object):
+class AutoCalibrator:
     """Automatically calibrate a spectrum by convolving it with a filter.
 
     A note on nomenclature: for historic reasons, 'channels' is used in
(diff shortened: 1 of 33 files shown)


markbandstra:
Wow, `pyupgrade` is really cool! We didn't know about it. Looks like the failures are due to `black`, so `black` should be run on the code. We would like to merge this in after your other recent PRs.

arunpersaud:
if you also use pre-commit, you can also add pyupgrade to it. That way each commit is checked (as long as the developer actually installs pre-commit)

markbandstra:
@jccurtis could you redo this with the current head?

jvavrek:
waiting on #244

jvavrek:
There are no active open PRs. Maybe now's a good time to re-do this, on a new branch to avoid all the conflicts?

markbandstra:
@jvavrek I agree. Best to redo this on a new branch. Do you have the spare cycles?

PR #262, opened 2021-04-11
setup: added pytest-black requirement

arunpersaud:
Setup.cfg has a --black option for tests, which I think needs pytest-black to be installed?


markbandstra:
Seems to make sense. Did this cause an error for you?

markbandstra:
Alternate solution: read `requirements-dev.txt` in `setup.py` in the same way we currently read `requirements.txt`, and send its contents to `tests_require`. I can work on this.

arunpersaud:
yes, I couldn't run "python3 -m pytest" since it complained about the --black option in setup.cfg

markbandstra:
Not sure the best workflow for making this contribution (could do a PR on your fork I suppose), but what I had in mind to make this PR more futureproof was:

* In `setup.py` add:
```python
with open("requirements-dev.txt", "r") as fh:
    REQUIREMENTS_DEV = fh.read()
```
* In `setup.py` change:
```python
tests_require=["pytest", "pytest-cov", "pytest-black"],
```
to
```python
tests_require=[_f for _f in REQUIREMENTS_DEV.split("\n")[1:] if _f],
```
* Optional: we could remove `flake8` and `pydocstyle` from `requirements-dev.txt` since they're no longer required.

What do you think?

arunpersaud:
sure, I can add this. requirements-dev also seems to include requirements.txt... should that be removed too or parsed out when loading the test-requirements? I'm not that familiar with requirements files

markbandstra:
> sure, I can add this. requirements-dev also seems to include requirements.txt... should that be removed too or parsed out when loading the test-requirements? I'm not that familiar with requirements files

Yes, good point, and my intention was to filter out that line with the `[1:]` in `REQUIREMENTS_DEV.split("\n")[1:]`, so that will return just the packages unique to `requirements-dev.txt`.

arunpersaud:
Just pushed an update. Let me know if this is what you were thinking

markbandstra (approved):
Perfect, thanks!

markbandstra:
@arunpersaud we've conventionally been letting the person who worked on the PR do the final update and merge commit, so please have a go at this one and #264 . I can also do it if you prefer.

PR #264, opened 2021-04-13
peakfinder: remove extra append to fwhms

arunpersaud:
The  fwhm gets appended a few lines below again. To me it seem as if it is a bug to also append outside of the if following if statement.


markbandstra (approved):
Wow, this is definitely a bug and needs to be fixed. Unclear to me how it's not been noticed before.

PR #265, opened 2021-04-20
Calibration fit statistics

jvavrek:
Fit stats for calibrations:
 - R^2
 - chi^2
 - dof
 - reduced chi^2

### DIFF
--- a/becquerel/core/calibration.py
+++ b/becquerel/core/calibration.py
@@ -406,6 +406,10 @@ class Calibration(object):
     def points_y(self):
         return self._points_y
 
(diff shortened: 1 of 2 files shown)


jvavrek:
@markbandstra any thoughts on this one? In addition to fit statistics, I also added a `plot()` method similar to what I added to `energycal.py` earlier. There are also some basic tests that we can expand upon if we add more tests of correctness before merging this into `main`.

markbandstra (approved):
Great addition, thanks for adding these features!

PR #266, opened 2021-04-23
update peakfinder to suppress double peaks and update plots

jvavrek:
TODO:
- [x] add a regression test: test that old peakfinder logic produces double peaks and new logic does not

### DIFF
--- a/becquerel/core/peakfinder.py
+++ b/becquerel/core/peakfinder.py
@@ -302,11 +302,12 @@ class PeakFinder(object):
     def plot(self, facecolor="red", linecolor="red", alpha=0.5, peaks=True):
         """Plot the peak signal-to-noise ratios calculated using the kernel."""
         bin_edges = self.spectrum.bin_edges_raw
+        bin_centers = self.spectrum.bin_centers_raw
 
         if facecolor is not None:
-            plt.fill_between(bin_edges[:-1], self.snr, 0, color=facecolor, alpha=alpha)
+            plt.fill_between(bin_centers, self.snr, 0, color=facecolor, alpha=alpha)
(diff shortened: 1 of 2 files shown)


becquerel/core/peakfinder.py
```diff
@@ -389,15 +390,9 @@ def find_peaks(self, xmin=None, xmax=None, min_snr=2, max_num=40):
...
-        d1 = (self.snr[2:] - self.snr[:-2]) / 2
-        d1 = np.append(0, d1)
-        d1 = np.append(d1, 0)
-        d2 = self.snr[2:] - 2 * self.snr[1:-1] + self.snr[:-2]
-        d2 = np.append(0, d2)
-        d2 = np.append(d2, 0)
+
         # find maxima
-        peak = (d1[2:] < 0) & (d1[:-2] > 0) & (d2[1:-1] < 0)
+        peak = (self.snr[2:] <= self.snr[1:-1]) & (self.snr[:-2] < self.snr[1:-1])
```
markbandstra, line 395: I was initially hesitant to remove the first and second derivative logic here since it incorporates information from another index away and thus might be smoother and less noisy, however since `snr` is already smoothed it's a moot point.

markbandstra: Optional, but what do you think about this ordering, I find it more logical:

`peak = (self.snr[:-2] < self.snr[1:-1]) & (self.snr[1:-1] >= self.snr[2:])`

jvavrek: Sure

markbandstra (approved):
Thanks, looks good!

PR #269, opened 2021-04-28
Run actions on schedule (nightly) and manually

jccurtis:
See title 😄

### DIFF
--- a/.github/workflows/coveralls.yaml
+++ b/.github/workflows/coveralls.yaml
@@ -4,6 +4,7 @@ on:
   push:
     branches: [ $default-branch ]
   pull_request:
+  workflow_dispatch:
 
 jobs:
   coveralls_test:
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -11,6 +11,10 @@ on:
   push:
     branches: [ $default-branch ]
   pull_request:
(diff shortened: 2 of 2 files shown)


jccurtis:
@markbandstra this fixes the issue with the coveralls default branch from earlier. Just needed to reset everything. This also adds nightly builds and manually triggered builds.

markbandstra:
> Just needed to reset everything.

Can you elaborate on this in case we run into the same issue in the future? Is it the addition of `workflow_dispatch:` to `coveralls.yaml`?

jccurtis:
@markbandstra For sure. I needed to login to coveralls.io with my github creds. 

Then I went to repos:

image

Then settings:

image

Then delete at the bottom of the page:

image

Then I went to add repo:

image

Then found becquerel and turned it on. (In the following screenshot it is already on):

image

This purged old coverage info but correctly picked up the default branch (`main`). I think I originally set this up before we changed the default branch. Technically coveralls claims this will be picked up upon a resync in the settings but that did not work for me. This [issue](https://github.com/lemurheavy/coveralls-public/issues/36) details a similar problem.

markbandstra:
@jccurtis Great, thanks for adding that information.

markbandstra (approved):
Looks great

PR #274, opened 2021-06-02
np.float and np.int are deprecated

### LINKED ISSUE #272: np.float is deprecated
I get a lot of these now:

```
DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

Should be an easy fix.

cosama:
Closes #272.

This fixes those to to `float` and `int`.

### DIFF
--- a/becquerel/core/fitting.py
+++ b/becquerel/core/fitting.py
@@ -429,7 +429,7 @@ class Fitter(object):
     @y_unc.setter
     def y_unc(self, y_unc):
         if y_unc is not None:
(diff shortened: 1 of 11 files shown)


markbandstra (approved):
Looks good

PR #276, opened 2021-06-18
small code cleanup: use default arguments of dict.pop() and get()

arunpersaud:


markbandstra (approved):
Thanks, that definitely simplifies the code!

PR #278, opened 2021-07-01
Test specific_activity prop against NRC values where possible

jvavrek:
I was double-checking a few things and thought these tests would be useful. As documented, we compare the specific activities returned by `bq.Isotope` against values from an NRC publication where possible. For most of the isomers, this isn't possible, so we just test against bq's current results as a sort of regression test.


markbandstra (approved):
Good catch

PR #279, opened 2021-07-14
Fix #257 - fitting plot color cycles

### LINKED ISSUE #257: Fitter.custom_plot fails if more than three model components
`Fitter.custom_plot` fails with a `IndexError: list index out of range` if there are more than three components to the model. The culprit is [this line](https://github.com/lbl-anp/becquerel/blob/main/becquerel/core/fitting.py#L853-L854):

```python
        colors = ["#1f78b4", "#33a02c", "#6a3d9a"]
```

A possible fix would be to extend the list like this:
```python
        colors = ["C0", "C2", "C4", "C5", "C6", "C7", "C8", "C9"]
```
where I left out `"C1"` and `"C3"` since they are similar to other colors used in the plot (the centroids and best fit, respectively).

markbandstra:
Still generates errors when trying to create a regression test though. @cosama please have a look.


cosama:
@markbandstra I added a test for multiple gaussians and also an example to the notebook: https://github.com/lbl-anp/becquerel/blob/fix-issue-257/examples/fitting.ipynb. I think this ended up quite a productive meeting.

tests/fitting_test.py
```diff
@@ -317,3 +316,41 @@ def test_with_roi(self, sim_high_stat):
             compare_counts(fitter)
         # fitter.custom_plot()
         # plt.show()
+
+
+@pytest.mark.parametrize("method", ["lmfit", "lmfit-pml", "minuit-pml"])
+def test_gauss_gauss_gauss_line(method):
```
jvavrek, line 322: I think just document this test a bit and then we're good to go.

jvavrek (approved):

markbandstra:
Oops, just saw the approval, but there was a change requested... what should we do?

jvavrek:
> Oops, just saw the approval, but there was a change requested... what should we do?

Hotfix or don't worry about it. It's minor.

PR #280, opened 2021-07-21
Implement a square root binning function

dhellfeld:
Simple util to generate square root bins.

### DIFF
--- a/becquerel/core/utils.py
+++ b/becquerel/core/utils.py
@@ -131,11 +131,13 @@ def sqrt_bins(bin_edge_min, bin_edge_max, nbins):
     Square root binning
 
     Args:
-      bin_edge_min (float): Minimum bin edge (must be greater than 0)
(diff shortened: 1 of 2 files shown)


becquerel/core/utils.py
```diff
@@ -124,3 +124,18 @@ def bin_centers_from_edges(edges_kev):
...
     centers_kev = (edges_kev[:-1] + edges_kev[1:]) / 2
     return centers_kev
+
+
+def sqrt_bins(bin_min, bin_max, nbins):
+    """
+    Square root binning
+
+    Args:
+      bin_min (float): Minimum bin edge (must be greater than 0)
```
jvavrek: >= 0, right? Also, maybe worth checking with an `assert`, otherwise we just get less-than-clear `nan`s

dhellfeld: Good call. Added.

dhellfeld:
@jccurtis Thoughts

Issue #65, opened 2017-05-24
Floating-point precision in exponential decays


bplimley:
`IsotopeQuantity.decays_from` and `IsotopeQuantity.from_decays` can suffer in precision if halflife >> time interval. This is because the calculations involve a `1 - np.exp(-decay_const * time)` kind of expression, and if `decay_const * time` is very small, the exponent is very close to 1, and the difference loses precision because of these two floating point quantities. See `np.spacing`.

Better precision could be achieved in some cases by a Taylor series approximation:
```
1 - exp(-lambda * t)
~= 1 - (1 - lambda * t)
~= lambda * t
```
The code could use `np.spacing` to check whether the series approximation or floating point has better precision.

This is relevant for long-lived isotopes like K-40. (I'm getting ~0.1% loss of precision for an hour's interval with K-40, which was causing tests to fail.)

jvavrek:
https://numpy.org/doc/stable/reference/generated/numpy.expm1.html

PR #281, opened 2021-09-01
use np.expm1 for hopefully better precision in decay calcs

jvavrek:
Just trying this out for #65 quickly. `np.expm1` works but so far I have not been able to improve the tests---maybe it's a `float` precision issue and not specifically the accuracy of the exponential function(s).


jvavrek:
I get 0.3% error in Th-232

jvavrek:
From the Th-232 test case:
```
coeff: 5.648098347704398e-15
1-exp: 5.662137425588298e-15
expm1: 5.648098347704382e-15
```
The `1-exp` definitely suffers from more accuracy loss than `expm1`. Even if we can't improve the test here, the `1-exp` pattern should generally be avoided.

jvavrek:
I think the roundoff error shows up in the `quantity_at` method:
```py
print(f'{np.exp(-dt * self.decay_const) :.15e}')
```
gives
```
9.999999999999943e-01
```
whereas wolfram alpha gives
```
9.99999999999994351901652295618e-01
```
and so on

jvavrek:
Using the [mpmath library](https://mpmath.org/doc/current/basics.html) I was able to narrow down the problem to (surprise, surprise) the subtraction of two large floats in computing the number of decays from initial and final atoms.
```
decays_mp = A0_mp - A1_mp  # mpmath
# mpf('1338260399.999996220686822980893623310176184366976369411076073521060728603480625388314873533635429211721')
```

```
decays_py = A0_py - A1_py  # numpy
# 1342177280.0
```

for a ratio of
```
decays_mp / decays_py
# mpf('0.9970816969871492838016323602897102461577835433009038948089432128226225104038606497011136514212408749203')
```

The minimal fix seems to be to use `mp` instead of `np` in `quantity_at`:
```py
return self._ref_quantities[quantity] * mp.exp(-dt * self.decay_const)
```
but we import mpmath we might as well use it throughout. I'll think of how to try/except the import, and use numpy (and warn) if necessary.

tests/isotope_qty_test.py
```diff
@@ -1,7 +1,15 @@
 import datetime
 from dateutil.parser import parse as dateutil_parse
 import copy
-import numpy as np
+
+try:
+    from mpmath import mp as xp
```
markbandstra, line 6: Is `mpmath` something we would want to include in `requirements.txt`?

jvavrek: That would be ideal. It has gotten a little messy trying to account for both numpy and mpmath

jvavrek:
@markbandstra pointed out that we might be able to just rewrite the problematic `decays_from` method:
```py
t0 = utils.handle_datetime(start_time, error_name="start_time")
t1 = utils.handle_datetime(stop_time, error_name="stop_time")
dt = (t1-t0).total_seconds()
return self.atoms_at(start_time) * (-np.expm1(-dt * self.decay_const))
```
but this seems to introduce _more_ test failures on main

alexmalins:
Hi all,

I recently discovered the `becquerel` package. I read this PR and issue #65 with interest due to tackling similar issues when writing one of my own [codes](https://github.com/alexmalins/radioactivedecay).

The solution I came up with uses [SymPy](https://www.sympy.org/en/index.html) arbitrary precision computations, which also employs the mpmath library on the backend. Empirically I found that carrying 320 significant figures of precision through the decay calculations was sufficient to ensure no numerical error in the result once cast back into a double-precision float. 

But as this was only empirical, it would benefit from checking further against the final solution you use for `becquerel` once it is ready. And also to check against the specific test cases you had issues with, as I may not have considered similar scenarios.

It would be great to open a discussion & share experiences on this given the overlap.

jvavrek:
Thanks @alexmalins, I'll take a look. Your radioactivedecay module looks pretty nice, as well!

jvavrek:
@markbandstra I thought about this a bit more and `uncertainties` is going to create conflicts with basically any arbitrary-precision math library. AFAIK we can't do a `ufloat` of `sympy` arbitrary-precision numbers or the reverse, a `sympy` object comprising the two values in the `ufloat`.

I think the only workaround would be to implement a "`USFloat`", basically just a class with a nominal value and standard deviation both defined as separate `sympy` objects... but then we would have to reimplement all the basic math _and_ error propagation. Not worth it, IMO.

There is a package called [maabara](https://github.com/c2man/maabara/blob/master/docs/user_guide.rst) (stale since 2014) that could potentially solve this.

Otherwise, I think the best thing to do here is to revert everything to numpy, replace `1-exp()` with `np.expm1()`, and call it a day.

markbandstra:
So in the end are there any tests that need to be adjusted?

jvavrek:
> So in the end are there any tests that need to be adjusted?

Not at this time. I think the last batch of failed tests was due to NNDC connection issues.

markbandstra (approved):
Looks good for now, and gives us food for thought for the future if we need to revisit this issue.

PR #282, opened 2021-09-03
Let Spectrum.from_listmode fill use bin_edges_kev instead of _raw

jvavrek:
This just adds an `is_cal` kwarg to `Spectrum.from_listmode` to use `bin_edges_kev` instead of `_raw`, and then adds several tests to make sure nothing inadvertently breaks.


becquerel/core/spectrum.py
```diff
@@ -581,7 +581,7 @@ def from_file(cls, infilename):
 
     @classmethod
     def from_listmode(cls, listmode_data, bins=None, xmin=None, xmax=None,
-                      **kwargs):
+                      is_cal=False, **kwargs):
```
markbandstra, line 584: Please document the new kwarg and update the docstring to refer not only to `bin_edges_raw`.

tests/spectrum_test.py
```diff
@@ -243,34 +243,37 @@ def test_negative_input(spec_data):
...
+        spec = bq.Spectrum.from_listmode(lmd, is_cal=is_cal, bins=log_bins)
     elif t == 'default':
-        spec = bq.Spectrum.from_listmode(lmd)
+        spec = bq.Spectrum.from_listmode(lmd, is_cal=is_cal)
     else:
         return t
 
-    if use_cal:
+    if apply_cal:
         cal = bq.LinearEnergyCal.from_coeffs({'m': TEST_GAIN, 'b': 0.0})
```
markbandstra, line 257: This should be  raising a `DeprecationWarning` -- should we take this as an opportunity to update to the new `Calibration`? Something like:

```python
cal = bq.Calibration.from_polynomial([0.0, TEST_GAIN])
```

markbandstra (approved):
Looks good, just address the docstring comment and optionally the deprecation comment.

tests/spectrum_test.py
jvavrek: I don't see the DeprecationWarning because this branch is way behind main. I'll resolve conflicts and check.

PR #285, opened 2021-09-27
Automatically deploy to pypi on new version tags

jccurtis:
The title says it all 😄 but this will need to be tested.


cosama (approved):
I will approve for now, and open an new issue for 3.9. I feel we should test and deploy that version too.

PR #289, opened 2021-09-30
Version v0.4.2 and pre-commit rules/CI integration

jccurtis:


jccurtis:
@cosama this _should_ fix the CD issues with multiple versions and adds some pre-commit checking which will make @micahfolsom smile 😄

cosama:
I am just wondering is it sufficient to just build a single python wheel for all of these different versions/platforms? How does pypi decide that? I like the precommit stuff.

jccurtis:
The `dist` is determined by the `wheel` command:

https://github.com/lbl-anp/becquerel/blob/main/.github/workflows/pypi.yaml#L32

which is currently set to `dist` and makes a `.whl` which is python3 minor version and OS and arch independent. There is a nice write-up of this I just read on [medium](https://medium.com/ochrona/understanding-python-package-distribution-types-25d53308a9a).

Previously we released with `python setup.py bdist ...` ([ref](https://github.com/lbl-anp/becquerel/blob/main/RELEASING.md?plain=1#L26)) which is outdated.

jccurtis:
Also I updated the release docs to reflect the new CD process.

jccurtis:
@cosama @jvavrek @markbandstra  any thoughts?

jvavrek:
@jccurtis to confirm, this only restricts the CD to one python version? Not the testing?

jccurtis:
> @jccurtis to confirm, this only restricts the CD to one python version? Not the testing?

Yes! That is correct 😄

jvavrek (approved):

.github/workflows/pypi.yaml
```diff
@@ -15,8 +15,8 @@ jobs:
 
     strategy:
       matrix:
-        os: [ 'macos-latest', 'ubuntu-20.04', 'windows-latest' ]
-        python-version: ['3.6', '3.7', '3.8', '3.9']
+        os: [ 'ubuntu-20.04' ]
+        python-version: [ '3.9' ]
```
markbandstra, line 19: What is the reason for dropping these OSs and older versions? Not saying we shouldn't, just thinking of the rationale going forward.

PR #291, opened 2021-10-05
Version 0.4.4 with pypi CD tag fix

jccurtis:
Just another small fix!!

🤣


jvavrek (commented):
These are the only places the version number needs to be changed?

PR #293, opened 2021-10-12
Redo pyupgrade

jvavrek:
redoing #261 after many subsequent PRs

I ran `pyupgrade --py36-plus` on source, tests, and the top level, then ran `black .`


jvavrek:
@jccurtis all tests pass!

PR #300, opened 2021-11-19
Update nudat2 to nudat3

dhellfeld:
NNDC updated from nudat2 to nudat3. Old requests seem to be failing, I think this may fix it.


dhellfeld:
@jccurtis @markbandstra Looks like this worked. Thoughts? A fix to this would be great, as it is breaking some tooling I have that uses this package.

markbandstra (approved):
Nice one, glad it was a simple fix!

PR #303, opened 2021-12-17
Fitter method to compute component-wise areas and their uncertainties

jvavrek:
image

```py
>>> fitter.calc_area_and_unc()
2002647.92714046+/-1415.3747870339978

>>> (fitter.y.sum(), np.sqrt(fitter.y.sum()))
(2002646, 1415.1487554317391)

>>> fitter.model.components
[<lmfit.Model: Model(gauss, prefix='gauss0_')>,
 <lmfit.Model: Model(gauss, prefix='gauss1_')>,
 <lmfit.Model: Model(line, prefix='line_')>]

>>> areas = {comp.name: fitter.calc_area_and_unc(component=comp) for comp in fitter.model.components}
{"Model(gauss, prefix='gauss0_')": 99936.80253652389+/-656.5995003915539,
 "Model(gauss, prefix='gauss1_')": 99861.92313776605+/-645.2239560410361,
 "Model(line, prefix='line_')": 1802849.20146617+/-1535.034662198673}

>>> sum([a.n for a in areas.values()])
2002647.92714046
```


becquerel/core/fitting.py
```diff
@@ -836,6 +837,61 @@ def _likelihood_residual(self, params, data, weights, **kwargs):
...
+        # Reformat best_values info so the gradient calculation can handle _calc_area
+        names = list(self.best_values.keys())
+        values = list(self.best_values.values())
+
+        # Compute the area under the curve
+        area = _calc_area(values, xvals=xvals, model=model, names=names)
+
+        # Compute the gradient with respect to the best fit parameters
+        grad = nd.Gradient(_calc_area)
+        g = np.atleast_2d(grad(values, xvals=xvals, model=model, names=names)).T
```
jvavrek, line 882: Something goes awry in these lines with `lmfit`. The `g` values differ between `lmfit` and `iminuit`, as does the order of the `names`/`values`. That probably throws a wrench into the `g * cov * g` multiplication, leading to the wrong uncertainty.

jvavrek: The issue is with `lmfit`:
```py
>>> list(fitter.result.best_values.keys())
['line_m',
 'line_b',
 'gauss1_amp',
 'gauss1_mu',
 'gauss1_sigma',
 'gauss0_amp',
 'gauss0_mu',
 'gauss0_sigma']

>>> list(fitter.params)
['gauss0_amp',
 'gauss0_mu',
 'gauss0_sigma',
 'gauss1_amp',
 'gauss1_mu',
 'gauss1_sigma',
 'line_m',
 'line_b',
 'gauss0_fwhm',
 'gauss1_fwhm']
```

We can fix this by using `fitter.param_names` (which works for both backends) and filtering out the `_fwhm` parameters:
```py
names = [name for name in self.param_names if self.params[name].vary]
values = [self.param_val(name) for name in names]
```

markbandstra (changes requested):
Please add test(s) of the new method.

jvavrek:
> Please add test(s) of the new method.

Yeah, you got me there.

jvavrek:
@markbandstra I've added tests of the new method. Something strange is going on with `pytest-black` in the CI, but tests pass locally.

jccurtis:
> @markbandstra I've added tests of the new method. Something strange is going on with either `black` or `pytest-black` in the CI, but tests pass locally.

@jvavrek that is really weird. I triggered the jobs again. This appears to be something in `pytest-black` which was fixed in: https://github.com/shopkeep/pytest-black/issues/42

jccurtis:
@jvavrek can you try `pytest-black>=0.3.10`

jccurtis:
@jvavrek also try `pytest<7.0.0`

jvavrek:
> @jvavrek also try `pytest<7.0.0`

This worked 🚀

Why was the CI installing 7.x when conda and pypi both only have 6.2.5?

jccurtis:
> > @jvavrek also try `pytest<7.0.0`
> 
> This worked 🚀
> 
> Why was the CI installing 7.x when conda and pypi both only have 6.2.5?
> 
> Bizarrely, it looks like it was not installing 7.x: https://github.com/lbl-anp/becquerel/runs/4723159061?check_suite_focus=true#step:5:196

Well I'll be scratching my head on that one ... thanks for posting #307 ... I'm sure this will be resolved by the plugin developers eventually

Issue #305, opened 2021-12-18
ErfModel fails unless coupled to a GaussModel


jvavrek:
The `guess()` method of `ErfModel` defines its guess for its `sigma` parameter by looking for an existing parameter called `gauss_sigma`:
```py
(f"{self.prefix}sigma", "expr", "gauss_sigma")
```

But if your peak is an `expgauss`, then it won't find anything, and it'll error out. Is there a more robust guess we can use?

Tagging @cosama

cosama:
That was not intended to be like that. We especially moved to a `GaussErfModel` to be able to couple the two. They were separated earlier in the development of the fitting module, but we thought it was odd to have it coupled. 

I would use a guess similar to the gaussian case:

```
sigma = xspan * width_ratio / 10.0
```

with an additional arguement to the `guess` function of something like `width_ratio=0.5`.

Side note: Not sure about that factor of 10. That seems kind of a random choice as well, it could (should?) just be absorbed into the width_ratio. If I hear with ratio I would not guess that it will be divided by 10 for some reason.

I can file a merge request later.

PR #306, opened 2021-12-18
Fix ErfModel

cosama:
Closes https://github.com/lbl-anp/becquerel/issues/305.

@jvavrek Check this out if it works for you. If you have a good test for this, based on the case where you run into this, it would probably be a good idea to add this.

### DIFF
--- a/becquerel/core/fitting.py
+++ b/becquerel/core/fitting.py
@@ -30,7 +30,6 @@ class FittingWarning(UserWarning):
 
 # -----------------------------------------------------------------------------
 # Base functions
-# TODO: Should these be replaced with lmfit.models???
 # -----------------------------------------------------------------------------
 
 
@@ -147,7 +146,6 @@ def _is_count_like(y):
 class ConstantModel(Model):
     def __init__(self, *args, **kwargs):
         super().__init__(constant, *args, **kwargs)
-        # TODO: remove this min setting?
         self.set_param_hint(f"{self.prefix}c", min=0.0)
 
     def guess(self, y, x=None, dx=None, num=2):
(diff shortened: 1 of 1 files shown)


becquerel/core/fitting.py
```diff
@@ -201,7 +201,7 @@ def guess(self, y, x=None, dx=None, center_ratio=0.5, width_ratio=0.5):
         msk = (x >= (mu - xspan * width_ratio)) & (x <= mu + xspan * width_ratio)
 
         # TODO: update this, minimizer creates NaN's if default sigma used (0)
```
jvavrek, line 203: @cosama can you just remove this an related TODOs?

jvavrek (approved):
Looks good to me. Just update any out-of-date TODOs about guess values and good to go!

markbandstra:
Didn't notice this at the time but we should do better on our commit messages -- a2aeb1e is "bump" and seems to be entirely empty (how did that happen?), and d43ce2f is "cosama" and a single-line change.

jvavrek:
> Didn't notice this at the time but we should do better on our commit messages -- [a2aeb1e](https://github.com/lbl-anp/becquerel/commit/a2aeb1ebd78452e5d543171556c891c44a5c7936) is "bump" and seems to be entirely empty (how did that happen?), and [d43ce2f](https://github.com/lbl-anp/becquerel/commit/d43ce2f5525f8e1762ae7f2a5d1e504b17eef20d) is "cosama" and a single-line change.

Those were just test commits to try re-triggering the frozen CI (though I suppose the messages could have said that, yes)

markbandstra:
Got it, thanks!

PR #310, opened 2022-01-15
Allow lmfit models

jvavrek:
Previously we could not directly use predefined `lmfit` models. We would get some strange type error in the `_guess_params_defaults` method. The cause can be shown here:
```py
# this is why we get strings unpacked!
for comp in lmfit.models.GaussianModel().components:
    guess = comp.guess(fitter.y_roi, x=fitter.x_roi, dx=fitter.dx_roi)
    for par in guess:
        print(par)
```
```
amplitude
center
sigma
fwhm
height
```
whereas `becquerel` was expecting a list like
```py
[("gauss_amp", "value", amp), ("gauss_amp", "min", 0), ("gauss_mu", "value", 511), ...]
```
so it was unpacking the single string (e.g. "amplitude") instead of the tuple of guess information. This can be avoided with the new `_parameters_to_bq_guess` function, which will generate this list of tuples.

Now we can do
```py
fitter = spec.fit(
    [bq.fitting.ConstantModel(prefix="constant"), lmfit.models.VoigtModel(prefix="voigt")]
)
```

However this raises the question of why this list of tuples was necessary in the first place, rather than an `lmfit.Parameters` object.


jvavrek:
Adding Models into the existing model-string-name-based fitting testing frameworks isn't working out, so I've added an example to the fitting notebook.

markbandstra:
Please add tests!

jvavrek:
We should get some `black` format failures again here.

jvavrek:
> Please add tests!

Trying to add say `lmfit.models.GaussianModel(prefix="gauss_")` to the `HIGH_STAT_SIM_PARAMS` is _impenetrable_, even after handling the base model params stuff near lines 167-170 in `fitting_test.py`. Something deep, deep within `becquerel` does not set parameters correctly and we end up calling a Gaussian with the wrong parameters, which ends up returning 0 for y and uncertainty across all of x.

As a result, I've added a single `test_lmfit_and_bq_models` that covers one specific case. In the general case, why our testing architecture needs to be so complex and yet so brittle to generate synthetic data, fit it, and compare it against known values is beyond me, as is why we need to translate so much of `lmfit` into `becquerel`'s own representations. Maybe a discussion for another time.

jccurtis:
> We should get some `black` format failures again here.

See #315 for a variety of fixes.

markbandstra (approved):
Strong work

PR #311, opened 2022-01-18
Downgrade fitting error

jvavrek:
Return 0 error estimate and warn instead of crashing in `Fitter.calc_area_and_unc()`

### DIFF
--- a/becquerel/core/fitting.py
+++ b/becquerel/core/fitting.py
@@ -887,12 +887,16 @@ class Fitter:
         grad = nd.Gradient(_calc_area)
         g = np.atleast_2d(grad(values, xvals=xvals, model=model, names=names)).T
 
+        # Compute the gradient with respect to the best fit parameters
+        grad = nd.Gradient(_calc_area)
+        g = np.atleast_2d(grad(values, xvals=xvals, model=model, names=names)).T
+
         # Compute the variance in the area estimate: Tellinghuisen Eq. 1
         if "minuit" in self.backend:
-            covariance = np.array(self.result.covariance)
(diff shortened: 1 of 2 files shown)


tests/fitting_test.py
```diff
@@ -334,10 +334,11 @@ def test_component_areas(self, sim_high_stat):
             covariance = np.array(fitter.result.covariance)
         else:
             covariance = np.array(fitter.result.covar)
-        # We can at least check that it properly errors, then skip the rest of the tests
+        # We can at least check that it properly warns, then skip the rest of the tests
         if not covariance.sum():
```
markbandstra, line 338: Please make sure this test is guaranteed to be evaluated.

becquerel/core/fitting.py
```diff
@@ -892,7 +893,11 @@ def _calc_area(param_vec, **kwargs):
         else:
             covariance = np.array(self.result.covar)
         if not covariance.sum():
```
markbandstra, line 895: Why not use `if np.isclose(covariance.sum(), 0):`?

markbandstra: Or is it returning `None`?

tests/fitting_test.py
markbandstra: Also see the `if not covariance.sum()` comment above.

becquerel/core/fitting.py
jvavrek: Yes, it was returning None, and then np.array(None).sum() was behaving strangely. Cleaned up in the latest commit!

markbandstra (approved):

PR #312, opened 2022-01-19
Add gauss + double exponential tail model

micahfolsom:
In working with some CZT-type detectors, I explored some advanced peak shapes, and found one that worked pretty well that's not included in lmfit. It's a Gaussian summed with exponential tails on either side, characterized by 3 parameters each. This PR will add this model to the list of built-ins, along with supporting docs and tests.


micahfolsom:
I added in the model with a reasonable `guess()` function and param hints, and was able to use it on my own (real, CZT) data successfully. It produced ~the same result as doing the same fit with lmfit by itself.

The tests don't pass, probably because the 2 cutoff parameters tend to not contribute much and can easily lead to misfits (I will often, but not always, fix these), though I'm not totally sure. The test is using the model to generate the test high stats data so it should "just work" with any model function, right?

jvavrek:
Looks like you're getting nan/inf results. Maybe you need to tighten up some parameter bounds in the fit?

micahfolsom:
This is shaping up - I wrote a new test to account for the 2 cutoff params that can be a little temperamental. In creating a notebook example, I ran into this issue: https://github.com/lbl-anp/becquerel/issues/317

Otherwise, I just need to finish writing documentation/comments and this should be ready for review.

jvavrek:
And for the split on first underscore, maybe something like:
```py
>>> s = "abc_123_xyz"
>>> "_".join([x for x in s.split("_")[1:]])
'123_xyz'
```

markbandstra:
Might be good to write a one-line function for clarity: `split_after_first_underscore` or something.

micahfolsom:
@jvavrek and @markbandstra there was a much easier solution: `maxsplit` kwarg to `str.split()`, nice and easy

micahfolsom:
This is ready to go @jvavrek, @jccurtis and @markbandstra if you get a chance to take a look

becquerel/core/fitting.py
```diff
@@ -72,6 +72,68 @@ def expgauss(x, amp=1, mu=0, sigma=1.0, gamma=1.0):
...
+        Right-side exponential tail cutoff, affects "length" of the tail
+
+    Returns
+    -------
+    numpy.ndarray
+        Y-data
+    """
+    alpha = -1.0 / (2.0 * sigma**2)
+    ltail_func = np.zeros(shape=(len(x),))
+    rtail_func = np.zeros(shape=(len(x),))
```
jvavrek, line 125: You could do `np.zeros_like(x)` here

jvavrek (approved):

markbandstra:
@micahfolsom can you update this branch? It looks like there are conflicts.

becquerel/core/fitting.py
micahfolsom: `zeros_like` is [a lot slower](https://github.com/numpy/numpy/issues/9909), and this is in a critical place, so I think we should stick with `zeros`

jccurtis: @micahfolsom what the ... 🤦 ... nice find.

markbandstra (approved):

PR #315, opened 2022-02-02
Update pre commit and black

jccurtis:
* Updates black to the recent stable release
* Adds additional tools to pre-commit:
    * check-json
    * check-merge-conflict
    * check-yaml
    * mixed-line-ending
    * requirements-txt-fixer (automatic fixes)
    * autoflake (automatic fixes)
    * flake8
    * markdownlint-cli (automatic fixes)
    * prettier (automatic fixes)
* Updates all files for compatibility with the new rules

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -1,0 +1,7 @@
+{
+  "default": true,
+  "MD013": { "code_blocks": false},
+  "MD024": {"allow_different_nesting": true},
+  "MD033": false,
+  "MD034": false
+}
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -1,6 +1,6 @@
 {
   "default": true,
-  "MD013": { "code_blocks": false},
+  "MD013": {"code_blocks": false},
   "MD024": {"allow_different_nesting": true},
   "MD033": false,
   "MD034": false
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -1,7 +1,7 @@
 {
   "default": true,
-  "MD013": {"code_blocks": false},
-  "MD024": {"allow_different_nesting": true},
+  "MD013": { "code_blocks": false },
+  "MD024": { "allow_different_nesting": true },
   "MD033": false,
-  "MD034": false
+  "MD034": false,
 }
--- a/becquerel/core/calibration.py
+++ b/becquerel/core/calibration.py
@@ -235,18 +235,18 @@ def _validate_expression(
     if len(param_indices) > 0:
         if param_indices.min() != 0:
             raise CalibrationError(
-                f"Minimum parameter index in expression is not 0:\n"
+                "Minimum parameter index in expression is not 0:\n"
                 f"{expression}\n{param_indices}"
             )
         if not np.allclose(np.diff(param_indices), 1):
             raise CalibrationError(
-                f"Parameter indices in expression are not contiguous:\n"
+                "Parameter indices in expression are not contiguous:\n"
                 f"{expression}\n{param_indices}"
             )
     if params is not None:
         if len(param_indices) != len(params):
             raise CalibrationError(
-                f"Not enough parameter indices in expression:\n"
+                "Not enough parameter indices in expression:\n"
                 f"{expression}\n{param_indices}"
             )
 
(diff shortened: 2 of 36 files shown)


becquerel/core/peakfinder.py
```diff
@@ -293,7 +293,7 @@ def add_peak(self, xpeak):
 
     def plot(self, facecolor="red", linecolor="red", alpha=0.5, peaks=True):
         """Plot the peak signal-to-noise ratios calculated using the kernel."""
-        bin_edges = self.spectrum.bin_edges_raw
+        self.spectrum.bin_edges_raw
```
jvavrek, line 296: This again looks like an autoformatter change. I'm not sure I like it... I would prefer a notification that a variable is unused instead.

becquerel/tools/df_cache.py
```diff
@@ -85,7 +85,7 @@ def write_file(self):
             raise CacheError("Cache has not been fetched or loaded")
         try:
             self.df.to_csv(self.filename, float_format="%.12f")
-        except:
+        except Exception:
```
jvavrek, line 88: More autoformatter changes? This is better than a bare `except`, but should we specialize the exception further?

tests/plotting_test.py
```diff
@@ -80,7 +79,7 @@ def test_plot_uncal_counts(uncal_spec):
     plt.close("all")
 
 
-def test_plot_uncal_counts(cal_spec):
+def test_plot_cal_counts(cal_spec):
     """Plot an calibrated spectrum"""
```
jvavrek, line 83:
```suggestion
    """Plot a calibrated spectrum"""
```

becquerel/core/calibration.py
```diff
@@ -235,24 +235,27 @@ def _validate_expression(
     if len(param_indices) > 0:
         if param_indices.min() != 0:
             raise CalibrationError(
-                f"Minimum parameter index in expression is not 0:\n{expression}\n{param_indices}"
+                f"Minimum parameter index in expression is not 0:\n"
+                f"{expression}\n{param_indices}"
```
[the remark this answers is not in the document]
jccurtis: Oh weird. Flake8 should catch that ...

becquerel/core/calibration.py
```diff
@@ -235,24 +235,27 @@ def _validate_expression(
...
             )
 
     # make sure the expression can be evaluated
     if params is not None:
         x_arr = np.linspace(domain[0], domain[1], num=n_eval)
         for x_val in x_arr:
             try:
-                y = _eval_expression(
+                _eval_expression(
```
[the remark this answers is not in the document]
jccurtis: Nope, it wasn't used below so I removed it (flake8)

becquerel/core/peakfinder.py
jccurtis: I am not following why you want a notification that a variable is unused. An unused var creates an extra reference in memory which is not used below so it doesn't make sense to track it. This is a flake8 error (pep something or another) and not removed by black

becquerel/tools/df_cache.py
jccurtis: Could do but I'm not sure what the exception should be. This maintains the same behavior and is a flake8 rule

becquerel/core/peakfinder.py
jvavrek: Right, we should just delete the line entirely instead.

.markdownlint.yaml
```diff
@@ -0,0 +1,7 @@
+{
+  "default": true,
+  "MD013": { "code_blocks": false},
```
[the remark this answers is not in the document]
jccurtis: 4872607cf9eacc7f74249b885c4e2dc8ffea8706

becquerel/core/peakfinder.py
jccurtis: Good point. I had originally read it as executing a function. Since this is a property which just returns a hidden attribute I've remove it: 20c093d501e0d6cb86bd41c0be6a3243be2bf2b9

jvavrek:
Now we just need NNDC to stop ruining the tests

.markdownlint.yaml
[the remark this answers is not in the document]
jccurtis: that is intentional because of this strange yaml formatting which markdown lint needs

jvavrek (approved):
Nice!

PR #318, opened 2022-02-10
Release v0.5.0

jccurtis:
Release new minor version. The features will be updated in the release notes on main when it is tagged

* bump version
* Move black to requirements

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,4 @@
 -r requirements.txt
-black>=22.1.0
 bump2version
 coverage
 flake8>=4.0.1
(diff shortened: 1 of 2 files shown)


jccurtis:
@markbandstra @jvavrek for some reason the reviewers and labels drop downs are not working ATM. This is ready for a look

jccurtis:
I changed the test action to make sure you can `import becquerel` before installing the dev reqs. This was a bug when importing the `calibration` on a normal install because `black` could not be imported (it was only a dev req).

markbandstra:
> I changed the test action to make sure you can `import becquerel` before installing the dev reqs. This was a bug when importing the `calibration` on a normal install because `black` could not be imported (it was only a dev req).

Good catch!

markbandstra (approved):

Issue #309, opened 2022-01-14
Incompatibility between calibration and parsers when calibrated energies can be negative


jvavrek:
Often we can end up with negative calibrated energies, e.g., in the first few bins if the calibration linear term is negative. These are clipped to zero by default in the new calibration module. However multiple zero edges will throw an error that the bin edges are not monotonically increasing.

One can normally change the clipping on the domain and range when doing calibrations manually, but not so with parsers, which automatically apply the calibration. One cannot change the clipping ranges passed into the calibration methods because the `DEFAULT_DOMAIN` etc are passed to those methods when the module is _imported_, not when the function is called.

We could either provide more kwargs to the parsers to prevent this, or allow negative energies (i.e., broaden or remove the clipping).

werthm:
I agree, this is an issue for me as well. DEFAULT_RANGE should allow negative energy values as well.

jvavrek:
I believe @markbandstra is working on this in #319

markbandstra:
Closed by #319

PR #319, opened 2022-02-16
Calibration raises warning when values are outside its range

### LINKED ISSUE #314: Calibration should raise a warning if it needs to clip to its range
A few times in practice a situation like the following has come up:

```python
import becquerel as bq
spec = bq.Spectrum.from_file("tests/samples/nai_detector.spe")
cal = bq.Calibration.from_linear([-10, 2.02])
spec.apply_calibration(cal)
```

And the result is that the `Spectrum` raises a `ValueError`:

```
ValueError                                Traceback (most recent call last)
Input In [22], in <module>
----> 1 spec.apply_calibration(cal)

File becquerel/becquerel/core/spectrum.py:1151, in Spectrum.apply_calibration(self, cal)
   1144     warnings.warn(
   1145         "The use of bq.EnergyCalBase classes is deprecated "
   1146         "and will be removed in a future release; "
   1147         "use bq.Calibration instead",
   1148         DeprecationWarning,
   1149     )
   1150 except AttributeError:
-> 1151     self.bin_edges_kev = cal(self.bin_edges_raw)
   1152 self.energy_cal = cal

File becquerel/becquerel/core/spectrum.py:535, in Spectrum.bin_edges_kev(self, bin_edges_kev)
    533     raise SpectrumError("Bad length of bin edges vector")
    534 elif np.any(np.diff(bin_edges_kev) <= 0):
--> 535     raise ValueError("Bin edge energies must be strictly increasing")
    536 else:
    537     self._bin_edges_kev = np.array(bin_edges_kev, dtype=float)

ValueError: Bin edge energies must be strictly increasing
```

The issue is that for the first few channels near zero, the calibration curve produces a negative output, and it implicitly clips it to the default range of `0` to `1e5`. So when the `Spectrum` sees that the first few channels have zero width in energy space, it raises this error.

This error is confusing because it doesn't point directly to the culprit -- that the calibration parameters need to be adjusted, or the calibration range needs to be extended -- and instead can lead to wasted time looking around `Spectrum`.

One solution would be to have `Calibration` raise an error if it goes out of range, but that outcome was considered and not favored when writing it because then fitting would become more complicated by often raising errors while searching for good parameter fits. Instead, it would be helpful for it to at least raise a warning so that this error can be more quickly understood, and for that warning to note that a `ValueError` might follow from `Spectrum`.

The range clipping is performed in this line:

https://github.com/lbl-anp/becquerel/blob/main/becquerel/core/calibration.py#L132

markbandstra:
Fixes #314. Now a warning is raised when a `Calibration` function yields values outside its specified range. For example:

```python
import becquerel as bq
spec = bq.Spectrum.from_file("tests/samples/nai_detector.spe")
cal = bq.Calibration.from_linear([-10, 2.02], rng=[0, 1e2])
spec.apply_calibration(cal)
```

Yields:

```
SpeFile: Reading file tests/samples/nai_detector.spe
/Users/msbandstra/Dropbox/Projects/Becquerel/becquerel/becquerel/core/calibration.py:157: CalibrationWarning: Function values are below the lower range (0.0) and above the upper range (100.0)
  warnings.warn(msg, CalibrationWarning)
```

The warning is raised but no longer are the values themselves clipped to the range as this has led to various problems.


becquerel/core/calibration.py
```diff
@@ -12,7 +13,7 @@
 from .. import io
 
 DEFAULT_DOMAIN = (0, 1e5)
```
jccurtis, line 15: Add comment (note) before this about the required monotonicity in this range

markbandstra:
Also fixes #309. Ready for review.

becquerel/core/spectrum.py
```diff
@@ -567,12 +567,13 @@ def bin_edges_raw(self, bin_edges_raw):
             self._bin_edges_raw = np.array(bin_edges_raw, dtype=float)
 
     @classmethod
-    def from_file(cls, infilename, verbose=False):
+    def from_file(cls, infilename, verbose=False, cal_kwargs={}):
```
jvavrek, line 570: Best to avoid mutable default params

becquerel/parsers/cnf.py
```diff
@@ -75,7 +75,7 @@ def _read_energy_calibration(data, index):
     return coeff
 
 
-def read(filename, verbose=False):
+def read(filename, verbose=False, cal_kwargs={}):
```
jvavrek, line 78: Mutable default params here too

tests/calibration_test.py
```diff
@@ -378,7 +395,15 @@ def test_calibration_fit_from_points(name, args):
...
-        cal2 = Calibration.from_points(args[0], points_x, points_y, params0=args[1])
+        cal2 = Calibration.from_points(
+            args[0],
+            points_x,
+            points_y,
+            params0=args[1],
+            domain=cal1.domain,
+            rng=cal1.range,
+        )
+        print(args[1])
```
jvavrek, line 406:
```suggestion
```

tests/calibration_test.py
```diff
@@ -466,7 +497,7 @@ def test_calibration_inverse():
 
     # evaluate the inverse for a value outside the range
     with pytest.raises(CalibrationError):
-        cal1.inverse(-10.0)
+        cal1.inverse(-2e6)
     with pytest.raises(CalibrationError):
         cal1.inverse(2e6)
```
jvavrek, line 502: Can you expand this to show that we can update the range of `cal1` and make the error disappear?

jvavrek (changes requested):

jvavrek:
I still kind of think that the default domain and range should be unbounded, especially given that the `domain` is really only used for validation and default plot x values. But that's maybe for another time.

markbandstra:
Not sure how to clear the "changes requested" status of this PR given that I have resolved all of the issues flagged by @jvavrek ...

jvavrek:
> Not sure how to clear the "changes requested" status of this PR given that I have resolved all of the issues flagged by @jvavrek ...

I think I just had to hit Approve. Send it! 🚀

Issue #325, opened 2022-03-28
Isotopes notebook has errors


jaromrax:
# Comment 1

```
activated_qty = IsotopeQuantity(iso2, date='2017-05-01', bq=103.2)
na23_qty = ni.activate(barns, initial=iso, activated=activated_qty)
```
gives nan g

# Comment 2

```
/home/ojr/.local/lib/python3.8/site-packages/becquerel/tools/isotope_qty.py:351: DeprecationWarning: bq_now() is deprecated and will be removed in a future release. Use bq_at(date=None) instead.
```
is shown many times....

# Comment 3

```
/usr/local/lib/python3.8/dist-packages/ipykernel/ipkernel.py:283: DeprecationWarning: `should_run_async` will not call `transform_cell` automatically in the future. Please pass the result to `transformed_cell` argument and any exception that happen during thetransform in `preprocessing_exc_tuple` in IPython 7.17 and above.
```
appears also sometimes in different notebooks.

# Comment 4

Also JSON from compendium may not be loaded with all materials...

jccurtis:
Hi @jaromrax, thanks for the notes! I've added some sections to your description to improve our ability to address each one.

1. Can you expand this to a MWE, i.e. what are the `iso2`, `barns`, `iso` and `activated_qty` vars?
2. See #326, this will go in soon. Thanks for the head's up!
3. I cannot find this text in any of our notebooks. I believe this was fixed here: https://github.com/ipython/ipykernel/pull/617. If you don't want to update `ipykernel` you can downgrade `ipython` to `7.10` as mentioned here: https://github.com/ipython/ipykernel/issues/540#issuecomment-699566981. FYI if you are using a `conda` env (highly recommended) you can update this with  `conda update ipykernel jupyter`.
4. @markbandstra is this related to #324?

markbandstra:
Yes, I'm guessing comment 4 pertains to #324, and I hope to get in a fix soon.

jccurtis:
Excellent, thanks @markbandstra!

jvavrek:
> # Comment 1
> ```
> activated_qty = IsotopeQuantity(iso2, date='2017-05-01', bq=103.2)
> na23_qty = ni.activate(barns, initial=iso, activated=activated_qty)
> ```
> 
> gives nan g

[...]

PR #331, opened 2022-03-31
Fix numpy complex dtype deprecation warning

jccurtis:
Found by @sfaaland


jccurtis:
@jvavrek the tests for the iso quantity are failing. I haven't dug in more but this might be related to #325. Either way the change in this PR is for fitting so we should be able to merge safely.

jvavrek:
I think NNDC is just down. I get similar errors locally on an older `main`

jvavrek (approved):

PR #333, opened 2022-04-01
Increase number of flaky reruns and add 2 s delay

### LINKED ISSUE #332: Flaky NNDC tests failing
We still have the flaky tests failing in the CI. See this [pipeline](https://github.com/lbl-anp/becquerel/runs/5779050487?check_suite_focus=true)

- [x] Add to dev requirements

jvavrek:
Closes #332


jvavrek:
@jccurtis I feel like my increasing the retries and adding a delay is ultimately a bad idea... this means up to 20 s _per test_, and if NNDC is down as it was in the last pipeline, the whole test suite can take an hour and a half to run!

jccurtis:
@jvavrek can we change the tests to use xdist and run in parallel?

markbandstra:
Maybe we should add a simple test (a `requests` POST?) and see if it is successful. If not, we can disable the NNDC tests. We also might have to do this for XCOM.

jvavrek:
With `pytest-xdist` we get some strange cache errors:
```sh
______________________ TestCacheExceptions.test_bad_path _______________________
[gw0] linux -- Python 3.6.15 /opt/hostedtoolcache/Python/3.6.15/x64/bin/python

self = <df_cache_test.TestCacheExceptions object at 0x7f084ab[57](https://github.com/lbl-anp/becquerel/runs/5825240685?check_suite_focus=true#step:7:57)9b0>

    def test_bad_path(self):
        """Test ExampleCache.check_path() exception for a bad path."""
        d = ExampleCache()
        d.path = "/bad/path"
        with pytest.raises(CacheError):
            d.check_path()
        with pytest.raises(CacheError):
>           d.check_file()
E           Failed: DID NOT RAISE <class 'becquerel.tools.df_cache.CacheError'>

tests/df_cache_test.py:67: Failed
```

@markbandstra can you comment? Are we perhaps getting a race condition here?

markbandstra:
I think that could be a race condition... it looks like setting `d.path` doesn't update `d.filename`, and it's possible that another test has created a file with the same name. I think that test could be fixed by the line

```python
        d.filename = "/bad/path/filename.csv"
```

right before that test.

jvavrek:
Now we have a `DeadlockError`: https://github.com/lbl-anp/becquerel/runs/5826003988?check_suite_focus=true

Possibly related:
- https://github.com/pytest-dev/pytest-xdist/issues/492
- https://github.com/ocrmypdf/OCRmyPDF/issues/456

jvavrek:
[Downgrading to `pytest<=5.3.5`](https://github.com/pytest-dev/pytest-cov/issues/398#issuecomment-628206557) almost worked, and then failed for python 3.9.

jvavrek:
> @jvavrek can we change the tests to use xdist and run in parallel?

In short, no. I've opened #336 to track it further and removed `xdist` from this PR. Ready for review!

cosama (approved):

PR #337, opened 2022-04-08
Add parallel pytest'ing to the test action

### LINKED ISSUE #336: Run tests in parallel with `pytest-xdist`
@jvavrek can we change the tests to use xdist and run in parallel?

_Originally posted by @jccurtis in https://github.com/lbl-anp/becquerel/issues/333#issuecomment-1087928469_

I was unable to get `pytest-xdist` to work properly—it seems to cause race conditions when used in conjunction with `pytest-cov`. Perhaps we can revisit this in the future. See the original post for further discussion on the original attempt.

jccurtis:
…  and assure each file's tests run on the same node.

Closes #336

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -42,5 +42,8 @@ jobs:
         run: python -c "import becquerel;"
       - name: Install dev dependencies
         run: pip install -r requirements-dev.txt
+      # Run tests on 4 node (processes) in parallel while guaranteeing that the
+      # tests from each file in tests/ runs in sequence on the same node. This
+      # is required for the code coverage function correctly.
       - name: Test with pytest
-        run: pytest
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yaml
```diff
@@ -42,5 +42,9 @@ jobs:
         run: python -c "import becquerel;"
       - name: Install dev dependencies
         run: pip install -r requirements-dev.txt
+      # Run tests on 4 node (processes) in parallel while guaranteeing that the
+      # tests from each file in tests/ runs in sequence on the same node. This
+      # is required for the code coverage function correctly.
```
jvavrek, line 47: Not just to function correctly, but to function without failing the pipeline!

jvavrek (approved):

PR #339, opened 2022-05-02
Fix stable IsotopeQuantity

jvavrek:
Fixes the sub-issue in #325 where we were getting `nan` for a mass of a sample. The issue was with how stable isotopes were handled in the various mass/activity/atoms conversions.

Coauthored with @cosama


examples/isotopes.ipynb
```diff
@@ -12,7 +12,12 @@
   {
    "cell_type": "code",
    "execution_count": 1,
-   "metadata": {},
+   "metadata": {
```
cosama, line 15: Can you clean up the notebook? I think you need to run something like:

```bash
jupyter nbconvert --ClearMetadataPreprocessor.enabled=True isotopes.ypynb
```

cosama (changes requested):

examples/isotopes.ipynb
jvavrek:
```
jupyter nbconvert --ClearMetadataPreprocessor.enabled=True --to notebook isotopes.ipynb
```

cosama (approved):

PR #351, opened 2022-08-23
peakfinder: convert exception to printing a warning. Fixes #350

### LINKED ISSUE #350: Exception during peak finding
I ran into a problem where I have a spectrum that has a pileup in its largest bin and the resulting self.snr is constant for the last few bins... this creates a second derivative that d2=0, which raises and exception. Perhaps it's better to just print a warning, so that the program can continue?
Screen Shot 2022-08-23 at 10 00 31 AM

blue: spectrum
orange: self.snr from peakfinder

I'll submit a pull request in a minute for a possible code change

arunpersaud:
Potential fix for  #350

### DIFF
--- a/becquerel/core/peakfinder.py
+++ b/becquerel/core/peakfinder.py
@@ -278,7 +278,11 @@ class PeakFinder:
                 / bw**2
             )
             if d2 >= 0:
-                raise PeakFinderError("Second derivative must be negative at peak")
+                warnings.warn(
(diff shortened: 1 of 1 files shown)


markbandstra (approved):
Thanks for the contribution!

markbandstra:
Looks like the tests no longer pass because this change results in a `PeakFinderError` no longer being raised for a particular case. Strangely this error is only seen in the coveralls test, and the error is opaque in the pip_build_tests.

PR #356, opened 2022-08-30
Check parameter names

### LINKED ISSUE #355: iminuit parameter lookup does not error on bad parameter names
image

This feels like a bug in iminuit, not becquerel, but we could certainly check the parameter name first.

jvavrek:
Closes #355 and fixes a few small related issues.


markbandstra (approved):

jvavrek:
I don't think the `FittingError` strings are formatted properly...

PR #357, opened 2022-10-05
changes all np.exp to np.expm1 in istope_qty

cosama:
This will fix the current main failing for `numpy==1.23.X` on ubuntu.

See also: https://github.com/numpy/numpy/issues/22386


markbandstra:
Previous related PR where we started using `expm1`: #281 

Original issue that prompted the use of `expm1`: #65

jvavrek (approved):

PR #361, opened 2022-11-02
Do not cancel other tests if there is a failure

markbandstra:
Some of the tests remain flaky, and by default all of the running tests get canceled if there is one failure. This change is an attempt to keep the tests running.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         os: ["macos-latest", "ubuntu-20.04", "windows-latest"]
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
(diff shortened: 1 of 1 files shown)


jvavrek (approved):

jvavrek:
Side note: should we remove python 3.6 from the tests matrix and replace it with 3.10 (and maybe 3.11)? 3.6 reached its EOL 10 months ago.

markbandstra:
Sounds good, I can add that change to this PR.

markbandstra:
Looks like it is testing against 3.10 now, but on the webpage it is still expecting 3.6 tests to pass. I think this is an artifact of the web frontend, so I am inclined to merge as it is.

jvavrek:
> Looks like it is testing against 3.10 now, but on the webpage it is still expecting 3.6 tests to pass. I think this is an artifact of the web frontend, so I am inclined to merge as it is.

Sounds good to me, go for merge

PR #364, opened 2023-01-11
Update code to pass correct POST requests to XCOM

markbandstra:
Their API must have changed slightly recently. This PR contains a small fix.


cosama (approved):
Looks good.

cosama:
@markbandstra This should be ready to be merged. Thanks for looking into this.

PR #368, opened 2023-04-04
add Fitter.param_rel_unc and adjust pandas append to concat

jvavrek:
Convenience function for fit parameter relative uncertainty

### DIFF
--- a/becquerel/core/fitting.py
+++ b/becquerel/core/fitting.py
@@ -1081,6 +1081,12 @@ class Fitter:
         else:
             raise FittingError("Unknown backend: {}", self.backend)
 
+    def param_rel_unc(self, param):
+        """
+        Relative error of fit parameter `param`
+        """
+        return self.param_unc(param) / self.param_val(param)
+
     @property
     def best_values(self):
         """Wrapper for dictionary of best_values."""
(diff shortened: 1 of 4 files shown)


markbandstra (approved):

jvavrek:
Only the 3.7 jobs are passing. Everything else gives
```
ERROR tests/isotope_qty_test.py::test_isotopequantity_from_comparison[Tl-208] - AttributeError: 'DataFrame' object has no attribute 'append'
```
Is the pandas dependency out of date?

jvavrek:
Locally with Python 3.9.13 all tests pass, but I do see the warning:
```
tests/wallet_cache_test.py::TestWalletCardCache::test_fetch
  /Users/jvavrek/becquerel/becquerel/tools/wallet_cache.py:64: FutureWarning: The frame.append method is deprecated and will be removed from pandas in a future version. Use pandas.concat instead.
    self.df = self.df.append(df_chunk)
```
So it's probably a newer version of `pandas` (maybe the [new version 2.0](https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html)?) compared to my local v1.5.1.

jvavrek:
@markbandstra all pipelines passed!

PR #372, opened 2023-06-09
try datetime format with all underscores

jvavrek:
`"%Y_%m_%d_%H_%M_%S"` is a non-standard but common-enough datetime format that `dateutil.parser.parse` can't handle but that we should support.

### DIFF
--- a/becquerel/core/utils.py
+++ b/becquerel/core/utils.py
@@ -100,7 +100,10 @@ def handle_datetime(input_time, error_name="datetime arg", allow_none=False):
         )
         return datetime.datetime(input_time.year, input_time.month, input_time.day)
     elif isinstance(input_time, str):
-        return dateutil_parse(input_time)
+        try:
+            return datetime.datetime.strptime(input_time, "%Y_%m_%d_%H_%M_%S")
+        except ValueError:
+            return dateutil_parse(input_time)
     elif input_time is None and allow_none:
         return None
     else:


markbandstra:
How has this come up? Can you add a regression test?

becquerel/core/utils.py
```diff
@@ -100,7 +101,10 @@ def handle_datetime(input_time, error_name="datetime arg", allow_none=False):
         )
         return datetime.datetime(input_time.year, input_time.month, input_time.day)
     elif isinstance(input_time, str):
-        return dateutil_parse(input_time)
+        try:
+            return dateutil_parse(input_time)
+        except ParserError:
+            return datetime.datetime.strptime(input_time, "%Y_%m_%d_%H_%M_%S")
```
markbandstra, line 107: If this fails, does it raise a useful error? If not we should also wrap this in a `try... except`.

jvavrek: Yes, if `strptime` fails due to an incorrect format that `dateutil_parse` also couldn't handle first, we get
```py
ValueError: time data '2021_01_01-12_01_01' does not match format '%Y_%m_%d_%H_%M_%S'
```

markbandstra: As discussed, please update the docstring, including the `Raises` part to reflect the current behavior.

tests/test_utils.py
```diff
@@ -26,3 +27,11 @@ def test_sqrt_bins():
     # edge_max < edge_min
     with pytest.raises(AssertionError):
         be = bq.utils.sqrt_bins(100, 50, n_bins)
+
+
+@pytest.mark.parametrize(
+    "timestamp", [datetime.date(year=2023, month=6, day=14), "2023_06_14_00_00_00"]
```
markbandstra, line 33: Please cover all the possible cases. Also test failures.

tests/utils_test.py
```diff
@@ -0,0 +1,60 @@
...
+    with pytest.raises(AssertionError):
+        be = bq.utils.sqrt_bins(100, 50, n_bins)
+
+
+@pytest.mark.parametrize(
+    "timestamp",
+    [
+        datetime.date(year=2023, month=6, day=14),
+        datetime.datetime(year=2023, month=6, day=14, hour=0, minute=0, second=0),
+        "2023_06_14_00_00_00",
```
markbandstra, line 37: Can you also include an ISO 8601 string, like "2023-06-14T00:00:00.000Z-0000"? That one would go through `dateutils` but not `strptime`.

markbandstra: Time zone might mess things up since the others aren't timezone-aware, so it probably needs to be omitted.

jvavrek: Good call, yeah. I just dropped it in the check but not in the input.

markbandstra (approved):

PR #374, opened 2023-06-30
Handle UNIX timestamps in `handle_datetime`

jvavrek:
Passing a UNIX timestamp like `1686726000.0` to `handle_datetime` currently gives an error, but we can parse it with `datetime.fromtimestamp`.

### DIFF
--- a/becquerel/core/utils.py
+++ b/becquerel/core/utils.py
@@ -3,6 +3,7 @@
 import datetime
 from dateutil.parser import parse as dateutil_parse
 from dateutil.parser import ParserError
+from numbers import Number
 from uncertainties import UFloat, unumpy
 import warnings
 import numpy as np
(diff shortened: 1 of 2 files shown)


jvavrek:
One test failure occurs on the CI but not my machine:
```
FAILED tests/utils_test.py::test_handle_datetime[1686726000.0] - AssertionError: assert datetime.datetime(2023, 6, 14, 7, 0) == datetime.datetime(2023, 6, 14, 0, 0)
```
This is almost certainly a timezone issue.

The other test failure is in `isotope_qty_test.py`, which I will investigate.

markbandstra:
Sorry @jvavrek , I accidentally canceled the tests while trying to rerun them, I didn't see you had just pushed.

jvavrek:
> Sorry @jvavrek , I accidentally canceled the tests while trying to rerun them, I didn't see you had just pushed.

No problem. I only updated from main, so they were bound to fail again anyway.

markbandstra (approved):

PR #378, opened 2023-07-27
adds ability to pass kwargs to guess_param_defaults in Fitter

cosama:
This quick fix, adds the ability to call 

```
fitter.guess_param_defaults(...)
```

before fitting a module to change the default parameters of the guess. Parameters can be passed to the various components simply by referencing them with their prefix. Calling the following code:

```
import becquerel as bq

model = bq.fitting.GaussModel

fitter = bq.Fitter(
    model,
    ...
)

fitter.guess_param_defaults(gauss_width_ratio=0.05, update=True)

fitter.fit(backend="lmfit-pml")
```
for example will use the proper `width_ratio` in the `model.guess(...)` function, instead of the default value as done right now.

The current behavior -- ignoring all kwarg arguments -- looks like a bug to me.

Additionally I also added the ability to the `ExpGauss` model to create a left tailing distribution, by selecting `gamma` as negative. So far we only were able to define right tailing distributions.


markbandstra:
Please add a test or two to cover these changes.

becquerel/core/fitting.py
```diff
@@ -385,7 +388,8 @@ def guess(self, y, x=None, dx=None, center_ratio=0.5, width_ratio=0.05):
             (f"{self.prefix}sigma", "value", sigma),
             (f"{self.prefix}sigma", "min", 0.0),
             (f"{self.prefix}gamma", "value", 0.95),
```
cosama, line 390: Need to make this a parameter

cosama: Done

markbandstra (approved):

PR #379, opened 2023-08-08
Fix nndc._request()

TannerDalen:
As of about two weeks ago, requesting data from .jsp sites (like https://www.nndc.bnl.gov/nudat3/sigma_searchi.jsp) for wallet card and decay data information began returning empty request errors: "NNDCRequestError: Request failed: "

image

After speaking with a Technology Analyst at Brookhaven, they concluded a few things. sigma_searchi.jsp provides a response object with "ok == True" and "status_code == 200". It's "reason" field is set to an empty string instead of "OK". In line 439, the 'resp.reason != "OK"' condition causes _request() to fail, despite NNDC returning data. Removing this condition fixes it and returns data successfully.

image

### DIFF
--- a/becquerel/tools/nndc.py
+++ b/becquerel/tools/nndc.py
@@ -436,7 +436,7 @@ class _NNDCQuery:
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore", category=ResourceWarning)
                 resp = session.post(self._URL, data=self._data, stream=False)
-            if not resp.ok or resp.reason != "OK" or resp.status_code != 200:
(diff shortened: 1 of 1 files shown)


markbandstra:
Wow, thanks @TannerDalen for digging into this! Some of us had started to notice this problem but hadn't looked into it. Glad it is a simple fix.

Hope you are getting good use out of becquerel, please let us know when you notice other bugs or have other features you would like to see!

markbandstra (approved):

TannerDalen:
Awesome, will do. Thanks!

PR #384, opened 2023-11-27
Address a FutureWarning from pandas

markbandstra:
This change removes the following warning that occurs when querying XCOM data:

```
becquerel/tools/xcom.py:403: FutureWarning: Passing literal html to 'read_html' is deprecated and will be removed in a future version. To read from a literal string, wrap it in a 'StringIO' object.
  tables = pd.read_html(self._text, header=0, skiprows=[1, 2])
```

### DIFF
--- a/becquerel/tools/xcom.py
+++ b/becquerel/tools/xcom.py
@@ -9,6 +9,7 @@ References:
 
 """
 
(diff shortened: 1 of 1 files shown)


jvavrek:
See also #381

PR #388, opened 2024-02-16
Pyupgrade via pre-commit

jvavrek:
Automatically run pyupgrade via pre-commit.

TBD: which minimum python version flag to use.


jvavrek:
See also #260, #261, #293

markbandstra (approved):
Great! I have a few things to add to pre-commit that I have come across recently, will put those in a PR or two.

PR #393, opened 2024-02-16
Add nbQA pre-commit hooks and fix issues

markbandstra:
Adds [nbQA](https://github.com/nbQA-dev/nbQA) to our pre-commit suite.


jvavrek (approved):
The notebook diffs are hard to read, but after a quick skim it looks like mostly whitespace, quote, and trailing zero changes (as we might expect). Approved!

PR #405, opened 2024-04-30
Add Spectrum.attenuate method

jvavrek:
Compute what a spectrum would look like after being attenuated by a given areal density of a given material, using the XCOM tools.

### DIFF
--- a/becquerel/core/spectrum.py
+++ b/becquerel/core/spectrum.py
@@ -8,7 +8,7 @@ from copy import deepcopy
 import numpy as np
 from uncertainties import UFloat, unumpy
 
-from .. import io, parsers
+from .. import io, parsers, tools
 from . import fitting, plotting
 from .rebin import rebin
 from .utils import EPS, bin_centers_from_edges, handle_datetime, handle_uncs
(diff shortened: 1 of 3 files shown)


becquerel/core/spectrum.py
```diff
@@ -983,6 +983,48 @@ def _mul_div(self, scaling_factor, div=False):
...
+            xp=xd["energy"],
+            fp=xd["total_w_coh"],
+        )
+        att = np.exp(-mass_att_coeff * areal_density_gcm2)
+        # the following is not (yet?) possible due to Spectrum.__mul__ limitations:
+        # return self * att
+
+        # instead, copy the pattern from Spectrum._mul_div:
+        if self._counts is not None:
+            data_arg = {"counts": self.counts * att}
```
markbandstra, line 1023: This statement leads to a `Spectrum` with noninteger counts, which we really shouldn't allow.

jvavrek: In general we do allow noninteger counts data:
```py
class Spectrum:
    def __init__(self, ...):
    """
    ...
    Args:
          counts: counts per bin. array-like of ints, floats or UFloats, if
            uncs is not provided all values must be positive
    ...
    """
```

markbandstra: Fair point. My worry here was that uncertainties would not be correctly estimated. I thought that no matter the type the counts are treated as if they were integers [when assigning uncertainties](https://github.com/lbl-anp/becquerel/blob/main/becquerel/core/spectrum.py#L113C1-L116C50):

```
          uncs (optional): iterable of uncertainty on the counts for each bin.
            If counts is given, is NOT a UFloat type, and uncs is not given,
            the uncertainties are assumed to be sqrt(N), with a minimum
            uncertainty of 1 (e.g. for 0 counts).
```

**However**, looking further into it, it looks like `self.counts` will in fact be a `uarray`, in which case `self.counts * att` is also a `uarray` with correctly-propagated uncertainties, and the [`handle_uncs` utility](https://github.com/lbl-anp/becquerel/blob/main/becquerel/core/utils.py#L48) would successfully detect this case and use the correct uncertainties, as indicated by this docstring.

markbandstra (approved):

jvavrek:
The pre-commit stage is complaining about
```sh
tests/spectrum_test.py:880:55: E251 unexpected spaces around keyword / parameter equals
tests/spectrum_test.py:880:57: E202 whitespace before '}'
tests/spectrum_test.py:880:57: E251 unexpected spaces around keyword / parameter equals
```
via `flake8` due to
```py
raise ValueError(f"Invalid {areal_density_gcm2 = }")
```
which is clearly incorrect. Ignoring.

PR #411, opened 2024-05-10
Spectrum.from_file: enable pathlib.Path

jvavrek:


markbandstra (approved):
Once we implement `ruff`, we can turn on a feature that checks that we're using `pathlib` and not `os` for filename stuff if we want.

PR #412, opened 2024-05-13
updates to Fitter.custom_plot

jvavrek:
1. option to plot without drawing the additional fit info panel
2. remove hardcoded margin settings
3. remove `savefname` kwarg
4. associated notebook and docstring updates


jvavrek:
Should we also get rid of the `title*` kwargs? They only appear in the following line
```py
# Set fig title
if title is not None:
    fig.suptitle(
        str(title), fontweight=title_fontweight, fontsize=title_fontsize
    )
```
which is normally left to the user to handle.

markbandstra:
> Should we also get rid of the `title*` kwargs? They only appear in the following line
> 
> ```python
> # Set fig title
> if title is not None:
>     fig.suptitle(
>         str(title), fontweight=title_fontweight, fontsize=title_fontsize
>     )
> ```
> 
> which is normally left to the user to handle.

Yes, let's leave it to the user, so long as they can access `fig`.

markbandstra (approved):

PR #413, opened 2024-05-13
Fitter.uncertainties dict

jvavrek:
`Fitter` already provides a `dict` of `best_values`, so this provides the corresponding `dict` of uncertainties.


jvavrek:
Tests are failing due to the `*_fwhm` parameter names, whose values are directly calculated from the corresponding `*_sigma` values. Should be an easy fix, just need to find it.

markbandstra (approved):

PR #414, opened 2024-05-13
Update relative uncertainties and related code

jvavrek:
- add type hints to fit-uncertainties-related things
- ensure that relative uncertainties are non-negative

### DIFF
--- a/tests/fitting_test.py
+++ b/tests/fitting_test.py
@@ -304,7 +304,7 @@ class TestFittingHighStatSimData:
             fitter.param_unc("bad_name")
         if "gauss_amp" in fitter.param_names:
             u = fitter.param_rel_unc("gauss_amp")
-            assert u is None or u < 0.01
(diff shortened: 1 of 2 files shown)


tests/fitting_test.py
```diff
@@ -304,7 +304,7 @@ def test_with_init(self, sim_high_stat):
             fitter.param_unc("bad_name")
         if "gauss_amp" in fitter.param_names:
             u = fitter.param_rel_unc("gauss_amp")
-            assert u is None or u < 0.01
+            assert u is None or (u < 0.01 and u >= 0)
```
markbandstra, line 307: What about `(0 <= u < 0.01)`?

jvavrek: I usually avoid that because it doesn't work for numpy arrays (although we're not dealing with an array here)

markbandstra (approved):

PR #416, opened 2024-05-29
Add liccheck

jvavrek:


jvavrek:
Dunno why the pre-commit hook is failing here...

markbandstra:
I reproduced this error on my local machine.

I noticed in the CI output (and on my machine) that the environment for `liccheck` is initialized but not installed. I suspect it has to do with the `language: system` setting, which according to the [`pre-commit` docs](https://pre-commit.com/#system):

> This hook type will not be given a virtual environment to work with – if it needs additional dependencies the consumer must install them manually.

Removing `language: system` from the config on my local machine leads to successfully executing `liccheck` but it raises an error:

```
pkg_resources.DistributionNotFound: The 'uncertainties>=3.0.3' distribution was not found and is required by the application
```

jvavrek:
@markbandstra I think the error you saw just means you don't have `uncertainties` installed (maybe on your machine, or maybe just in that local environment). Trying the CI with that line removed now...

jvavrek:
Dunno, feels like a bug. Specifying `uncertainties>=3.0.3` should be no different than `black>=22.3.0` or `numba>=0.47.0`, which both work on the CI.

markbandstra:
You're right, it's looking in the virtual environment, which does not have `uncertainties` installed. Looking at `setuptools.pkg_resources`, it seems like `uncertainties` is only getting blamed here because of [this line in pkg_resources in the WorkingSet.resolve method](https://github.com/pypa/setuptools/blob/main/pkg_resources/__init__.py#L809):

```python
        requirements = list(requirements)[::-1]
```

so the last requirement in the from `requirements.txt` is the first one examined and found to be missing.

So it seems like we want the system's `liccheck` to run and check the system's environment, where the relevant libraries would be installed, so `language: system` should be correct after all. But how to fix the CI, I do not know. It seems like `pip` installing `requirements-dev.txt` should have done the trick.

I think maybe you didn't see it on your local machine because you probably ran `pip install liccheck` whereas I know I did not, thinking it would only need to exist in a virtual environment.

jvavrek:
Possibly because there's no `pip install requirements.txt` (or `-dev`) in the separate `pre-commit` job: https://github.com/lbl-anp/becquerel/blob/main/.github/workflows/pre-commit.yaml

jvavrek:
https://github.com/lbl-anp/becquerel/actions/runs/9304327492/job/25608628469?pr=416

```
Successfully installed ... uncertainties-3.1.7 ...
```

```
pkg_resources.DistributionNotFound: The 'uncertainties>=3.0.3' distribution was not found and is required by the application
```

🤷

markbandstra (approved):
Looks good, good idea!

jvavrek:
Now our `pre-commit / run-pre-commit` job passes but the automatic `pre-commit.ci - pr` job fails?

jvavrek:
As we continue to modernize project configuration, we'll probably move requirements into the `pyproject.toml`... in that case, we can simplify the list of licenses via regex.

PR #417, opened 2024-05-29
Enable pylint and other linting rules in ruff

markbandstra:
This PR enables the following [rulesets](https://docs.astral.sh/ruff/rules/) in [`ruff`](https://docs.astral.sh/ruff/):

  - `SIM`: [flake8-simplify](https://pypi.org/project/flake8_simplify/) rules to simplify code
  - `PL`: [pylint](https://pypi.org/project/pylint/) rules to "check[] for errors, enforce[] a coding standard, look[] for [code smells](https://martinfowler.com/bliki/CodeSmell.html), and [] make suggestions about how the code could be refactored"
  - `FURB`: [refurb](https://pypi.org/project/refurb/), "for refurbishing and modernizing Python codebases"

I thought these made useful improvements and would be helpful for maintaining our code going forward.

### DIFF
--- a/becquerel/core/fitting.py
+++ b/becquerel/core/fitting.py
@@ -693,7 +693,7 @@ class Fitter:
         raise FittingError(f"Unknown model type: {m}")
 
     def _make_model(self, model):
-        if isinstance(model, str) or isinstance(model, Model):
+        if isinstance(model, (str, Model)):
             model = [model]
         # Convert the model(s) to a list of Model classes / Model instancess
         self._model_cls_cnt = {}
@@ -750,10 +750,9 @@ class Fitter:
 
     def guess_param_defaults(self, update=False, **kwargs):
         defaults = self._guess_param_defaults(**kwargs)
-        if update:
-            if defaults is not None:
-                for dp in defaults:
-                    self.set_param(*dp)
+        if update and defaults is not None:
+            for dp in defaults:
+                self.set_param(*dp)
         return defaults
 
     def fit(self, backend="lmfit", guess=None, limits=None):
@@ -871,9 +870,7 @@ class Fitter:
                         min_vals[lim[0]] = lim[2]
                     elif lim[1] == "max":
                         max_vals[lim[0]] = lim[2]
-                limits_i = {
-                    p: (min_vals.get(p, None), max_vals.get(p, None)) for p in free_vars
-                }
+                limits_i = {p: (min_vals.get(p), max_vals.get(p)) for p in free_vars}
             except NotImplementedError:
                 # If the model/component does not have a guess() method
                 limits_i = {}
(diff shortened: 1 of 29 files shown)


becquerel/core/fitting.py
```diff
@@ -1440,8 +1435,8 @@ def custom_plot(
             # Add info about the ROI and units
             if self.roi:
                 s += "ROI: [{:.3f}, {:.3f}]\n".format(*self.roi)
-            s += "X units: {:s}\n".format(self.xmode if self.xmode else "None")
-            s += "Y units: {:s}\n".format(self.ymode if self.ymode else "None")
+            s += "X units: {:s}\n".format(self.xmode or "None")
+            s += "Y units: {:s}\n".format(self.ymode or "None")
```
jvavrek, line 1439: I dislike this change---it relies on a subtlety unique to python that not even many intermediate python programmers know, just for the sake of a little brevity. Could we disable this?

becquerel/parsers/iec1455.py
```diff
@@ -95,7 +95,7 @@ def read(filename, verbose=False, cal_kwargs=None):
     energy_channel_pairs = []
     energy_res_pairs = []  # currently unused
     energy_eff_pairs = []  # currently unused
-    with Path(filename).open() as f:
+    with Path(filename).open(encoding="utf-8") as f:
```
jvavrek, line 98: is this really necessary?

tests/fitting_test.py
```diff
@@ -31,7 +31,7 @@ def sim_data(x_min, x_max, y_func, num_x=200, binning="linear", **params):
         edges = np.linspace(x_min, x_max, num_x, dtype=float)
     elif binning == "sqrt":
         edges = np.linspace(np.sqrt(x_min), np.sqrt(x_max), num_x)
-        edges = edges**2
+        edges **= 2
```
jvavrek, line 34: this exists? TIL

jvavrek: not sure I like it

becquerel/core/fitting.py
markbandstra: Your alternate idea: convert both to `f"{str(self.xymode)}"` for maximum conciseness.

markbandstra: This is almost certainly [FURB110](https://docs.astral.sh/ruff/rules/if-exp-instead-of-or-operator/)

markbandstra:
I will redo this PR without implementing `ruff`'s [`preview` mode](https://docs.astral.sh/ruff/preview/) since it includes unstable rules.

becquerel/parsers/iec1455.py
markbandstra: This is [`PLW1514`](https://docs.astral.sh/ruff/rules/unspecified-encoding/).

PR #419, opened 2024-10-08
Handle an apparent XCOM bug in energies list

markbandstra:
The first energy in the list of user-specified energies seems to get discarded by the server. This is a kludge with a noisy warning to remind us to remove it when it gets fixed.

This kludge also works on the [XCOM website query](https://physics.nist.gov/PhysRefData/Xcom/html/xcom1.html) --- just add a "dummy" energy at the top of the "Additional energies in MeV" list.

### DIFF
--- a/becquerel/tools/xcom.py
+++ b/becquerel/tools/xcom.py
@@ -9,6 +9,7 @@ References:
 
 """
 
+import warnings
 from collections.abc import Iterable
 from io import StringIO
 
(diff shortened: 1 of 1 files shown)


markbandstra:
Apparently our tests have consistently been failing to due XCOM-related errors since August 1.

markbandstra:
The one failure is because I re-enabled `pre-commit.ci` to see if its behavior would change, but it still fails on `liccheck`. It has been turned off, so all relevant checks have passed.

jvavrek (approved):

PR #420, opened 2024-10-15
Update Th-232 series notebook example

markbandstra:
The [NNDC data querying example notebook](https://github.com/lbl-anp/becquerel/blob/main/examples/nndc.ipynb) needed two updates:

* It was using a deprecated `pandas` method (`DataFrame.append`), replaced with `pandas.concat`
* The radiation from the thorium series started at Th-228 for reasons obscure to me (maybe because we had a Th-228 source in the lab), so if you refer to it for the natural thorium background lines it is missing some important ones.


markbandstra:
@jvavrek seems like the only problem is a `liccheck` issue and I'm not sure why it's even checking the `pillow` version since it's not a dependency.

```console
Run Python License Checker...............................................Failed
- hook id: liccheck
- exit code: 255

gathering licenses...
45 packages and dependencies.
check authorized packages...
44 packages.
check unknown packages...
1 package.
    pillow (11.0.0): ['CMU License (MIT-CMU)']
      dependency:
          pillow << matplotlib
```

jvavrek:
CMU is fine, so you could add exactly `CMU License (MIT-CMU)` to the `.liccheck.ini`

We could also disable checking the downstream dependencies in the tool.

PR #427, opened 2025-02-22
Spectrum properties

jvavrek:
I noticed some opportunities to add useful spectrum properties and clean up `Spectrum.__str__` in the process.


jvavrek:
I believe the issue with testing std_devs here is that in the spectrum, bins with 0 counts are considered 0 ± 1. But in the gross counts or cps calculation, those bins add 0 counts and 0 uncertainty. So summing the ufloats and grabbing the std dev is not quite correct, we should do the error propagation ourselves.

markbandstra (approved):

PR #428, opened 2025-02-22
Run pre-commit autoupdate

markbandstra:
...and fix minor issues that have arisen.

### DIFF
--- a/becquerel/parsers/cnf.py
+++ b/becquerel/parsers/cnf.py
@@ -298,9 +298,7 @@ def read(filename, verbose=False, cal_kwargs=None):
     # clean up null characters in any strings
     for key, value in data.items():
         if isinstance(data[key], str):
-            data[key] = value.replace("\x00", " ")
-            data[key] = data[key].replace("\x01", " ")
-            data[key] = data[key].strip()
+            data[key] = value.replace("\x00", " ").replace("\x01", " ").strip()
 
     # create an energy calibration object
     if cal_kwargs is None:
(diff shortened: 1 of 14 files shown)


jvavrek:
why do we keep `prettier` around? 😆

markbandstra:
True, I could use this PR to drop it.

becquerel/parsers/cnf.py
```diff
@@ -296,9 +296,9 @@ def read(filename, verbose=False, cal_kwargs=None):
     data["counts"] = counts
 
     # clean up null characters in any strings
-    for key in data:
+    for key, value in data.items():
         if isinstance(data[key], str):
-            data[key] = data[key].replace("\x00", " ")
+            data[key] = value.replace("\x00", " ")
             data[key] = data[key].replace("\x01", " ")
             data[key] = data[key].strip()
```
jvavrek, line 304: I think this can be left as is

jvavrek: Or better yet, chained together

becquerel/parsers/spc.py
```diff
@@ -392,9 +392,9 @@ def read(filename, verbose=False, cal_kwargs=None):
         raise BecquerelParserError("Calibration parameters not found") from exc
 
     # clean up null characters in any strings
-    for key in data:
+    for key, value in data.items():
         if isinstance(data[key], str):
-            data[key] = data[key].replace("\x00", " ")
+            data[key] = value.replace("\x00", " ")
             data[key] = data[key].replace("\x01", " ")
             data[key] = data[key].strip()
```
jvavrek, line 399: chain together

jvavrek (approved):

PR #433, opened 2025-07-22
handle infinite relative uncertainty

jvavrek:
`lmfit` [recently switched](https://github.com/lmfit/lmfit-py/pull/1009) to returning `float` relative uncertainty values rather than `np.float64`, causing `ZeroDivisionErrors` to be raised rather than giving `np.inf` for expressions that look like `1.0 / 0.0`. We should preserve the IEEE-754-like behavior here of returning infinity since some of our downstream packages rely on it and `lmfit` does not want to revert.

### DIFF
--- a/becquerel/core/fitting.py
+++ b/becquerel/core/fitting.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import inspect
 import warnings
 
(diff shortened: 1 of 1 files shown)


jvavrek:
The code at issue is:
```py
    def param_rel_unc(self, param):
        """
        Relative error of fit parameter `param`
        """
        if self.param_unc(param):
>           return self.param_unc(param) / self.param_val(param)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E           ZeroDivisionError: float division by zero
```

which is actually out-of-date after https://github.com/lbl-anp/becquerel/pull/414. This unrelated fix coincidentally should cast everything to a numpy float type and give `np.inf` instead of the error.

So we need to create a new becquerel release and make sure to import that version downstream.

PR #437, opened 2025-11-12
Fixes the CNF reader

cosama:
Looks like there have been some changes in the reader. I updated the current implementation to be closer to the one in xylib here: https://github.com/wojdyr/xylib/blob/master/xylib/canberra_cnf.cpp

@alihanks Can you test this?


becquerel/parsers/cnf.py
```diff
@@ -3,14 +3,9 @@
 The CNF data format is not readily available to build a custom parser so this
 code is based on the CNF parsing code in the open source xylib project:
     https://github.com/wojdyr/xylib
-
-The relevant code is in these files:
-    https://github.com/wojdyr/xylib/blob/master/xylib/util.cpp
-    https://github.com/wojdyr/xylib/blob/master/xylib/canberra_cnf.cpp
```
markbandstra, line 9: It would be good to keep this information.

becquerel/parsers/cnf.py
```diff
@@ -19,64 +14,92 @@
...
 
-def _convert_date(data, index):
-    """Convert 64-bit number starting at index into a date."""
-    # 'data' is the number of 100ns intervals since 17 November 1858, which
-    # is the start of the Modified Julian Date (MJD) epoch.
-    # 3506716800 is the number of seconds between the start of the MJD epoch
-    # to the start of the Unix epoch on 1 January 1970.
-    d = _from_little_endian(data, index, 8)
-    t = (d / 10000000.0) - 3506716800
-    return datetime.datetime.utcfromtimestamp(t)
```
markbandstra, line 35: It would be good to keep this function or at least keep all of the comments explaining what the time conversion is doing.

becquerel/parsers/cnf.py
```diff
@@ -19,64 +14,92 @@
...
 
-def _convert_time(data, index):
-    """Convert 64-bit number starting at index into a time."""
-    # 2^(64) - number of 100ns intervals since timing started
-    d = _from_little_endian(data, index, 8)
-    d = (pow(2, 64) - 1) & ~d
-    return d * 1.0e-7
+    # Bitwise inversion (One's Complement) on 64-bit unsigned
+    val_inv = val ^ 0xFFFFFFFFFFFFFFFF
+    return val_inv * 1.0e-7
```
markbandstra, line 48: Likewise here, the information that the clock is 100ns is lost and buried in a magic number. The new function is nice in handling more cases, but has stripped some context.

becquerel/parsers/cnf.py
```diff
@@ -19,64 +14,92 @@
...
 
 def _from_pdp11(data, index):
-    """Convert 32-bit floating point in DEC PDP-11 format to a double.
+    """Convert 32-bit PDP-11 float to double."""
+    if index + 4 > len(data):
+        return 0.0
 
-    For a description of the format see:
-    http://home.kpn.nl/jhm.bonten/computers/bitsandbytes/wordsizes/hidbit.htm
-    """
```
markbandstra, line 51: Keep this comment

becquerel/parsers/cnf.py
```diff
@@ -19,64 +14,92 @@
...
-            return np.nan
-        else:
-            return 0.0
+        return np.nan if sign == -1 else 0.0
+
     h = (
-        data[index + 2] / 256.0 / 256.0 / 256.0
-        + data[index + 3] / 256.0 / 256.0
+        data[index + 2] / 16777216.0
+        + data[index + 3] / 65536.0
```
markbandstra, line 67: These changes hide context for the calculations in the magic numbers. There should be an explanation added.

PR #446, opened 2026-04-02
moves to pyproject.toml

### LINKED ISSUE #443: Migrate fully to pyproject.toml instead of setup.py
We should completely use the latest packaging standards and stop using `setup.py`.

Related to #441.

cosama:
Closes #443.

This has a quite a few changes:

* Move everthing that can be to `pyproject.toml`. That includes `requirements.txt`, `requirements-dev.txt` and `.liccheck.ini`
* It removes the full copyright text from `__metadata__.py`. The two files `LICENSE.txt` and `COPYRIGHT.txt` are copied in `pyproject.toml` to the package and will be present. I think that approach is way simpler, but still covers us sufficiently.
* `setup.py` is kept as legacy. It is not required in any way, but having it means `python setup.py` installs still work.
* `setuptools` as part of `dev` requirements is pinned to `<81`. That is because liccheck is not keeping up with modern packaging.


jvavrek (approved):

markbandstra (approved):

cosama:
All test seem to pass

